PR #103, opened 2024-03-29
Local Postgres

kraftp:
Update the quickstart to provide instructions on using DBOS Transact with an installed Postgres server instead of Docker.

### DIFF
--- a/docs/getting-started/quickstart.md
+++ b/docs/getting-started/quickstart.md
@@ -203,7 +203,7 @@ If successful, the script should print `Database started successfully!`.
 </TabItem>
 </Tabs>
 
-Next, let's create some tables in our database by running a schema migration:
+Next, let's create some tables in your database by running a schema migration:
 
    ```bash
 npx dbos migrate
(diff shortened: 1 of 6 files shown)


docs/getting-started/quickstart.md
```diff
@@ -35,100 +35,175 @@ export NVM_DIR="$HOME/.nvm"
...
-npx -y @dbos-inc/create -n <app-name>
+npx -y @dbos-inc/create@latest -n <app-name>
    ```
 
 Application names should be 3 to 30 characters long and contain only lowercase letters and numbers, dashes, and underscores.
 
 This creates a folder for your application, configures its layout, and installs required dependencies.
 If successful, it should print `Application initialized successfully!`.
-By default, `init` instantiates a "Hello, Database!" application which greets users and tracks the count of greetings per user.
+By default, it instantiates a "Hello, Database!" application which greets users and tracks the count of greetings per user.
```
qianl15, line 114: Maybe we can point people to two different next steps: 
- Running Locally
- Deploying to DBOS Cloud
This tutorial does seem to be longer than before..

qianl15 (approved):

PR #117, opened 2024-05-05
Update TypeORM Documentation

kraftp:
Documentation for https://github.com/dbos-inc/dbos-transact/pull/419

### DIFF
--- a/docs/tutorials/using-typeorm.md
+++ b/docs/tutorials/using-typeorm.md
@@ -6,7 +6,9 @@ description: Learn how to create and register TypeORM entities and perform trans
 
 ## TypeORM Overview
 [TypeORM](https://typeorm.io) is a popular TypeScript ORM.
-DBOS supports it as an alternative to [Knex](https://knexjs.org/) for transactional data management.
+It is based on the idea of creating [`Entity`](https://typeorm.io/entities) classes to represent each database table, with the persistent and join key fields marked with [decorators](https://typeorm.io/decorator-reference).
+Once entity classes are defined, TypeORM provides methods for storing, updating, and querying the entities via the [`EntityManager`](https://typeorm.io/working-with-entity-manager).
+DBOS supports TypeORM as an alternative to [Knex](https://knexjs.org/) for transactional data management.
 
 ## Usage
 DBOS supports essentially direct use of TypeORM, but a few additional steps are necessary to inform DBOS about the TypeORM entity list, manage database schemas, and use a transaction in the context of a workflow.
--- a/docs/tutorials/using-typeorm.md
+++ b/docs/tutorials/using-typeorm.md
@@ -5,29 +5,31 @@ description: Learn how to create and register TypeORM entities and perform trans
 ---
 
 ## TypeORM Overview
-[TypeORM](https://typeorm.io) is an ORM for TypeScript.  It is based on the idea of creating [`Entity`](https://typeorm.io/entities) classes to represent each database table, with the persistent and join key fields marked with [decorators](https://typeorm.io/decorator-reference).  Once entity classes are defined, TypeORM provides methods for storing, updating, and querying the entities via the [`EntityManager`](https://typeorm.io/working-with-entity-manager).  TypeORM can also be used to create and maintain the database schema.
-
-If you are using TypeORM, DBOS needs to know about it for inserting workflow status updates along the transactions used for application code.
-
-Use of TypeORM is optional.  DBOS supports several other libraries for transactional data management.
+[TypeORM](https://typeorm.io) is a popular TypeScript ORM.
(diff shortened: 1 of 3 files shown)


docs/tutorials/using-typeorm.md
```diff
@@ -105,41 +108,66 @@ If preferred, it is possible to define a `type` to clean up the transaction meth
...
   connectionTimeoutMillis: 3000
-  app_db_client: 'typeorm'
+  app_db_client: typeorm
+  migrate:
+    - npx typeorm migration:run -d dist/datasource.js
+  rollback:
+    - npx typeorm migration:revert -d dist/datasource.js
+runtimeConfig:
+  entrypoints:
+    - dist/src/operations.js
```
qianl15, line 136: Do we need these lines?

docs/tutorials/using-typeorm.md
```diff
@@ -53,24 +55,25 @@ class KVOperations {
...
-    await testRuntime.createUserSchema();
+TypeORM can also [automatically generate migration files from changes to your entity files](https://typeorm.io/migrations#generating-migrations).
+Assuming you started with our template, you can use this capability by running:
+
+```bash
+npx typeorm migration:generate -d dist/datasource.js migrations/<migration-name>
 ```
 
-Additional options for triggering schema migration during the application deployment process may be available in the future.
+This automatically generates a migration file containing commands to transition your database from its current schema to the schema defined in your entity files.
```
qianl15, line 76: Put a pointer about the `dist/datasource.js` file here

docs/tutorials/using-typeorm.md
qianl15: Okay this wasn't very obvious. Because we added a "datasource.ts" file, now the compiled operations file is no longer under `dist/operations.js`

docs/tutorials/using-typeorm.md
```diff
@@ -5,30 +5,32 @@ description: Learn how to create and register TypeORM entities and perform trans
 ---
 
 ## TypeORM Overview
-[TypeORM](https://typeorm.io) is an ORM for TypeScript.  It is based on the idea of creating [`Entity`](https://typeorm.io/entities) classes to represent each database table, with the persistent and join key fields marked with [decorators](https://typeorm.io/decorator-reference).  Once entity classes are defined, TypeORM provides methods for storing, updating, and querying the entities via the [`EntityManager`](https://typeorm.io/working-with-entity-manager).  TypeORM can also be used to create and maintain the database schema.
+[TypeORM](https://typeorm.io) is a popular TypeScript ORM.
+DBOS supports it as an alternative to [Knex](https://knexjs.org/) for transactional data management.
```
qianl15, line 9: I think it's good to put a pointer to entity manager here. It clarifies what we use in TypeORM for transactions.

docs/tutorials/using-typeorm.md
```diff
@@ -105,41 +108,66 @@ If preferred, it is possible to define a `type` to clean up the transaction meth
...
+const [dbosConfig, ] = parseConfigFile();
+
+const AppDataSource = new DataSource({
+    type: 'postgres',
+    host: dbosConfig.poolConfig.host,
+    port: dbosConfig.poolConfig.port,
+    username: dbosConfig.poolConfig.user,
+    password: dbosConfig.poolConfig.password as string,
+    database: dbosConfig.poolConfig.database,
+    ssl: dbosConfig.poolConfig.ssl as TlsOptions,
```
chuck-dbos, line 157: This is very clumsy.

docs/tutorials/using-typeorm.md
kraftp: The entities and migrations, which also need to be compiled, have the same effect.  We probably need to improve documentation around entrypoints later...

docs/tutorials/using-typeorm.md
chuck-dbos: ORMEntities decorator would be preferable to me, because the entity list is code, not configuration.  Certainly not both things, with them getting out of sync with each other.

kraftp: How does that get the info to the migrations though?  They explicitly want a path to a datasource file.

chuck-dbos: Sometimes it seems like TypeORM is very close but not quite there... so I'd be willing to write a little bit of code if that's what it takes to help the developer succeed.

kraftp: What are you suggesting?

docs/tutorials/using-typeorm.md
kraftp: Restored the original text

chuck-dbos (approved):

PR #123, opened 2024-05-10
Clarify sleep units

chuck-dbos:


docs/getting-started/quickstart-programming.md
```diff
@@ -199,7 +199,7 @@ export class Greetings {
             ctxt.logger.info(
                 "Press control + C to interrupt the workflow..."
             );
-            await ctxt.sleep(1);
+            await ctxt.sleepms(1000);
```
qianl15, line 202: I think we should keep the original `await ctxt.sleep(1)` and add a comment to it `// Sleep for 1 second`

qianl15 (approved):

PR #127, opened 2024-05-29
Doc change to no longer use `dashboard launch`

apoliakov:


maxdml (approved):
Should we add the `delete` doc here too?

apoliakov:
I don't believe there is currently any reason to run `dashboard delete`, unless you are moving to a different org. so let's put that there, when it happens...

PR #130, opened 2024-06-05
Doc for configured instances and AWS lib

chuck-dbos:


docs/tutorials/configured-instances.md
```diff
@@ -0,0 +1,99 @@
...
+description: Learn how to make workflows, transactions, and communicators reusable and configurable
+---
+
+In this guide, you'll learn how to make your DBOS functions configurable using instances.
+
+## Concepts
+Basic DBOS transactions, communicators, and workflows are just functions - they accept input parameters, apply those parameters to the database, or use them to place calls to external services.
+
+However, it is sometimes desirable to have configuration information available to DBOS functions.  Using function parameters for items such as access URLs, API keys, port numbers, and so on is a bad idea, so these are generally stored as configuration data in `dbos-config.yaml` or otherwise accessed from context.
+
```
kraftp, line 13: but another set of credentials (typo)

docs/tutorials/configured-instances.md
```diff
@@ -0,0 +1,99 @@
...
+If a function needs more than one configuration, such global settings are not adequate.  For example, an email-sending function may send email with one set of addresses and credentials for promotional materials to prospects, or another set of credential for replies to support inquiries from existing customers.
+
+### Instances
+Configured class instances are the DBOS Transact mechanism for creating multiple function configurations.  Rather than having `static` class member functions, configured instances are TypeScript object instances with non-static member functions that can access configuration information through `this`.
```
kraftp, line 17: They're TypeScript classes, no?

docs/api-reference/contexts.md
```diff
@@ -155,52 +157,56 @@ The [Koa Context](https://github.com/koajs/koa/blob/master/docs/api/context.md)
...
 To invoke workflows, use [`invokeWorkflow`](#handlerctxtinvokeworkflow) or [`startWorkflow`](#handlerctxtstartworkflow) instead.
 The syntax for invoking function `fn` in class `Cls` with argument `arg` is:
 
 ```typescript
 const output = await handlerCtxt.invoke(Cls).fn(arg);
 ```
 
-You don't need to supply a context to the invoked function&#8212;the DBOS Transact runtime does this for you.
-You can optionally provide a UUID idempotency key to the invoked function.
+You aren't to supply a context to the invoked function&#8212;the DBOS Transact runtime does this for you.
```
kraftp, line 171: Minor style point, but I liked the old way better here

docs/api-reference/communicatorlib.md
```diff
@@ -72,3 +72,69 @@ This communicator is provided in the `@dbos-inc/communicator-random` package.
...
+application:
+  aws_config:
+    aws_region: ${AWS_REGION}
+    aws_access_key_id: ${AWS_ACCESS_KEY_ID}
+    aws_secret_access_key: ${AWS_SECRET_ACCESS_KEY}
+```
+
+The default configuration section of `application.aws_config` is used for any communicator that has not been specifically configured.
+
+Individual AWS services can override this, for example the SES communicator uses `aws_ses_configuration` to specify the configuration(s) for use by SES:
```
kraftp, line 99: At least to me, all these defaults are a bit confusing and un-discoverable (how would I know it's called `aws_ses_configuration`?), especially as all the SES examples specify a particular named configuration.

docs/api-reference/communicatorlib.md
```diff
@@ -72,3 +72,69 @@ This communicator is provided in the `@dbos-inc/communicator-random` package.
...
+```yaml
+application:
+  aws_ses_configuration: aws_config_ses
+  aws_config_ses:
+    aws_region: ${AWS_REGION}
+    aws_access_key_id: ${AWS_ACCESS_KEY_ID_SES}
+    aws_secret_access_key: ${AWS_SECRET_ACCESS_KEY_SES}
+```
+
+In the event that there are multiple access keys for the same service, the application must be involved in determining the number and purpose of the configurations.  The application can do this with direct names, or indirect names.
```
kraftp, line 109: What's a direct name versus an indirect name?

docs/api-reference/communicatorlib.md
```diff
@@ -72,3 +72,69 @@ This communicator is provided in the `@dbos-inc/communicator-random` package.
...
+  aws_config_ses_admin:
+    aws_region: ${AWS_REGION}
+    aws_access_key_id: ${AWS_ACCESS_KEY_ID_SES_A}
+    aws_secret_access_key: ${AWS_SECRET_ACCESS_KEY_SES_A}
+```
+
+The application code will then have to specify which configuration to use when initializing the communicator:
+```typescript
+    // Initialize once ...
+    const sesCfg = configureInstance(SendEmailCommunicator, 'userSES', {awscfgname: 'aws_config_ses_user'});
```
kraftp, line 125: All the examples do this, defeating the point of all the complex default rules somewhat. Maybe just scratch all these defaults and require it always be specified?

docs/api-reference/contexts.md
chuck-dbos: The "need to" implied you could.  How about "You don't supply"?

docs/api-reference/communicatorlib.md
chuck-dbos: Yes, so none of these are discoverable.  aws_config, aws_config.aws_region, aws_config.aws_access_key_id, aws_config.aws_secret_access_key, aws_ses_configuration, not discoverable and not error checked; you could potentially "discover" them from the README or cut+paste an example.
What is the proposal to make them discoverable?  That a library has a way to extend the application schema section for the config file in some way?

docs/api-reference/communicatorlib.md
chuck-dbos: Easy enough to add another example

docs/tutorials/configured-instances.md
chuck-dbos: What is the proposed wording?

docs/api-reference/communicatorlib.md
chuck-dbos: I added another example in the README.md for SES (in the S3 PR), which is one place it belongs.  I will do something here too.

docs/api-reference/contexts.md
kraftp: Yeah that's much better!

docs/api-reference/communicatorlib.md
kraftp: My proposal is to get rid of them (because they're undiscoverable) and require the name always be supplied in `configureInstance`.

docs/tutorials/configured-instances.md
kraftp: "Configured instances are TypeScript classes with non-static member functions..."

kraftp (approved):

PR #156, opened 2024-07-10
fix(workflow-tutorial): add a 'y' to 'an'

robotnoises:
### Description

I was reading the docs this morning and noticed a small typo. This PR aims to fix it. My goal is to contribute the smallest change in the entire history of this project. 😤

> Please note: I don't know if this was the original intent of the author. It's possible that they want to make it  "...DBOS attempt to recover _a_ workflow..."


apoliakov (approved):
@robotnoises thank you! No contribution is too small! :)

qianl15:
Thanks! I think the original intention was indeed "...DBOS attempt to recover a workflow..."

PR #159, opened 2024-07-16
Update contexts.md

weisisheng:
fixed some misc. typos


apoliakov (approved):
Thanks Vince!

qianl15:
Actually, they are not typos -- `TransactionContext<T>` indicates it accepts a generic type parameter `T`, which is explained in the following paragraph.

apoliakov:
@qianl15 looks like the <T> was removed erroneously but Vince also fixed a typo in "transactioncontextt"

apoliakov:
Ah I think I see the intent. The misspelling "transactioncontextt" is purposeful. Sorry about that!

PR #160, opened 2024-07-17
Update logging.md

weisisheng:
Added some reference links.

### DIFF
--- a/docs/tutorials/logging.md
+++ b/docs/tutorials/logging.md
@@ -52,3 +52,5 @@ You can also configure the logging level as a CLI argument to the runtime:
 ```shell
 npx dbos start --loglevel debug
 ```
+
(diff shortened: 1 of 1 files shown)


apoliakov:
Hi @weisisheng thanks as always!

This isn't quite the right link though. We do not use PG log levels in this context. We use the Winston package with the levels as described here:
https://github.com/winstonjs/winston?tab=readme-ov-file#logging-levels
(See a further link from there to [RFC5424](https://tools.ietf.org/html/rfc5424))

*But* we also limit the levels. Per the top of the page, we only support `ctxt.logger.info()`, `...debug()`, `...warn` and `...error`. Generally the app developer is free to use these four methods as they see fit and then adjust this config as they wish to either filter out or expose more messages. 

Does this make more sense? Would you like to adjust your PR? Or if you like I can take a stab at consolidating....

weisisheng:
Thanks.  My mistake.  Let me look over winston and see if there is any need to edit/supplement.

qianl15:
Close this PR for now. Welcome to create a new PR for this issue!

PR #175, opened 2024-08-12
Update the docs for the new static analysis plugin version, 3.3.0

CaspianA1:


docs/api-reference/static-analysis.md
```diff
@@ -275,3 +276,8 @@ static async depositWorkflow(ctxt: WorkflowContext, data: TransactionHistory) {
 ```
 
 *Any global variable defined outside the scope of the workflow which is directly modified will result in a warning.*
+
+3. Malformed transactions:
+- Your transactions must have a `TransactionContext<T>` as the first parameter, where `T` is a supported database client.
```
kraftp, line 281: Be more clear as to what this actually means--we're looking for transactions that don't touch the database. The details don't matter.

kraftp (approved):

PR #178, opened 2024-08-14
Clarify / add missing cloud management docs

maxdml:


docs/cloud-tutorials/application-management.md
```diff
@@ -39,6 +39,10 @@ When you deploy an application, DBOS Cloud first runs [`npx dbos migrate`](../ap
 It then starts your application.
 After your application is deployed, DBOS Cloud hosts it at this URL, which is also printed by the deploy command:
 
+:::info
+Your application must not consume more than 1GB of disk space.
```
CaspianA1, line 43: Just curious - why is this a requirement? Wouldn't this detract some customers from using DBOS?

CaspianA1: And isn't this requirement repeated below?

maxdml: Just our current configuration. Can evolve in the future.

docs/cloud-tutorials/best-practices.md
```diff
@@ -0,0 +1,87 @@
...
+* Provisioning instances is an asynchronous process. You can check the status of the instance with `npx dbos-cloud db status <database-instance-name>`
+
+### Linked databases (BYODB)
+* DBOS Cloud will not configure `approle` for linked database
+* DBOS cloud will not provide a default SSL certificate for linked database [FIXME: what is the expected flow here?]
+
+## Users and organizations
+
+### Registering a new user
+- A new user needs to be signed up with DBOS cloud (through auth0) before registering a cloud account.
```
CaspianA1, line 78: 'cloud' should be capitalized here

docs/cloud-tutorials/application-management.md
chuck-dbos: Your application is allowed to consume "disk space"?!  What even is that?

maxdml: "must not use more .." updated

docs/cloud-tutorials/application-management.md
```diff
@@ -39,6 +39,10 @@ When you deploy an application, DBOS Cloud first runs [`npx dbos migrate`](../ap
 It then starts your application.
 After your application is deployed, DBOS Cloud hosts it at this URL, which is also printed by the deploy command:
 
+:::info
+Your application must not use more than 1GB of disk space.
```
qianl15, line 43: Rephrase it to something like: "Each app is allocated with 1GB of disk space." so it sounds more positive.

docs/cloud-tutorials/best-practices.md
```diff
@@ -0,0 +1,87 @@
+---
+sidebar_position: 8
+title: Best practices
+description: Tips and tricks for succesfully using DBOS Cloud
+---
+
+This section dives into specific behaviors and corner cases you might encounter when using DBOS Cloud.
```
qianl15, line 7: Mention "best practices" so it sounds more positive.

In general, I hope this docs can guide people into the right place in a more positive tone.

docs/cloud-tutorials/best-practices.md
```diff
@@ -0,0 +1,87 @@
...
+
+## Managing applications
+
+### Registration
```
qianl15, line 11: Since we're merging registration with deploy from the CLI, we can fold this into the deployment section.

docs/cloud-tutorials/best-practices.md
```diff
@@ -0,0 +1,87 @@
...
+* An application must be registered before being deployed. DBOS Cloud CLI will handle registration by default when you deploy an application for the first time.
```
qianl15, line 12: We won't need this soon.

docs/cloud-tutorials/best-practices.md
```diff
@@ -0,0 +1,87 @@
...
+* Only one request at a time can manipulate an application and you will get an `application is busy` error if you attempt multiple concurrent modifications.
```
qianl15, line 13: This is true for both deploy and delete. We can turn this section into a "General" section.

docs/cloud-tutorials/best-practices.md
```diff
@@ -0,0 +1,87 @@
...
+* Application names are unique within an [organization](#organizations-and-users-management)
+
+### Deployment
+* Interrupting a deployment request - for example, pressing control + C - will *not* cancel the deploy action on DBOS Cloud
+* DBOS Cloud will prune all development dependencies. Make sure to explicitly install any runtime dependency you need with `npm install`
+* Your application must not use more than 1GB of disk space
```
qianl15, line 19: Same suggestion as above, rephrase it to something like: "Each app is allocated with 1GB of disk space." so it sounds more positive.

Also mention the RAM and CPU limits, so people will know what to expect.

docs/cloud-tutorials/best-practices.md
```diff
@@ -0,0 +1,87 @@
...
+* DBOS Cloud stores each version of your application. You can download a specific version using the DBOS Cloud CLI [TOLINK WHEN MERGED]
+
+### During runtime
+* Nothing on the filesystem is persisted between deployments
```
qianl15, line 23:
```suggestion
* Files on disk are ephemeral. Nothing on the filesystem is persisted between deployments
```

docs/cloud-tutorials/best-practices.md
```diff
@@ -0,0 +1,87 @@
...
+* DBOS Cloud will keep around VMs running old versions of a workflow. This means, you need to be mindful of non-backward compatible database schema changes.
+
+### Deletion
+* Deleting an app is an asynchronous process and there is a small delay between a deletion request completing and the application being fully deleted. This means you could hit an "application is busy" error when re-deploying an application you just deleted
+* An application will be deleted once all PENDING workflows -- including these previous versions -- are completed
+* All application archives -- including previous versions -- will be deleted
+* You can elect to drop the application database alongside deleting an application. This will also delete all the database roles associated with the application in the database instance.
```
qianl15, line 30: The app role is always deleted when you delete an app. Also, I don't think we need to mention it at all as the role is transparent to users.

docs/cloud-tutorials/best-practices.md
```diff
@@ -0,0 +1,87 @@
...
+
+### Time travel
+You cannot enable time travel on an existing application. You will need to register a new application.
+See more information in the [time travel](#time-travel) section.
+
+### Swapping database instance
+While an application cannot change database once registered, you can change the database instance (server) hosting the application database. When doing so, you can also decide to redeploy a previous version of the application. This is how DBOS Cloud support point in time recovery today.
```
qianl15, line 37: Provide a link to the disaster recovery doc?

docs/cloud-tutorials/best-practices.md
```diff
@@ -0,0 +1,87 @@
...
+
+## Application databases
+
+:::tip
+Application databases are "logical" postgres databases. They are not to be confused with database *instances*, the server hosting application databases.
```
qianl15, line 42:
```suggestion
Application databases are databases within a Postgres database instance (server).
```

docs/cloud-tutorials/best-practices.md
```diff
@@ -0,0 +1,87 @@
...
+:::
+
+* Application databases cannot be shared across applications
+* An application cannot change its database once deployed
```
qianl15, line 46: Add some pointers for people who want to modify the database: delete the app and re-create it with a different name.

docs/cloud-tutorials/best-practices.md
```diff
@@ -0,0 +1,87 @@
...
+
+### Database roles
+
+* DBOS Cloud will create the following roles for each application:
+    * `dbosadmin` - a privileged role used to execute your schema migration steps during application deployment
+    * `approle` - a user-provided role you provide when provisioning the database instance. This role has
```
qianl15, line 52: The name of this role can either be user provided, or by default `dbos_user` (https://github.com/dbos-inc/dbos-transact/pull/598)

docs/cloud-tutorials/best-practices.md
```diff
@@ -0,0 +1,87 @@
...
+        * a CREATE grant on the application database
```
qianl15, line 53: I'm thinking we should probably just grant this role a superuser role, as it is meant to be used by our users for manual debugging and migrations.

docs/cloud-tutorials/best-practices.md
```diff
@@ -0,0 +1,87 @@
...
+        * an ALL PRIVILEGES grant on all schemas and [base tables](https://www.postgresql.org/docs/current/infoschema-tables.html#INFOSCHEMA-TABLES) in the database instance, expect for the schemas `pg_catalog`, `information_schema`, `dbos` or `pg_toast` schemas
+        * ownership on all base tables in the database instance
+    * `dbosproxyrole` - a read-only role used during [DBOS Time Travel](cloud-tutorials/timetravel-debugging)
+
+### Schema migrations
+* You can instruct DBOS Cloud to run [migration commands](api-reference/configuration#database) during application deploy. These are handled by `dbosadmin`. You can use them to install extensions, create tables, etc.
+* For more complex operations, you can use the DBOS Cloud CLI to connect to your database and run migrations manually.
+
+### System databases
+* DBOS Transact relies on a _system database_ to store metadata about the application. This database cannot be configured when running on DBOS Cloud.
```
qianl15, line 64:
```suggestion
* DBOS Transact relies on a _system database_ to store metadata about the application. This database name is `<app_db_name>_dbos_sys` and cannot be configured when running on DBOS Cloud.
```

docs/cloud-tutorials/best-practices.md
```diff
@@ -0,0 +1,87 @@
...
+### Registering a new user
+- A new user needs to be signed up with DBOS Cloud (through auth0) before registering a Cloud account.
+
+### Refresh tokens
+The preferred way to programmatically authenticate with DBOS Cloud is to use refresh tokens. Note they:
+* Expire after a year or a month of inactivity
+* Can be revoked manually
+
+## Time travel
+* Time travel will result in the creation of a new application database in the database instance
```
qianl15, line 86:
```suggestion
* Time travel will result in the creation of a new database called `<app_db_name>_dbos_prov` in the database server. 
```
Provide a link to: https://docs.dbos.dev/api-reference/system-tables#provenance-tables

docs/cloud-tutorials/best-practices.md
maxdml: I think we should keep it around and mention its done automatically for you.

docs/cloud-tutorials/best-practices.md
maxdml: I wanted to mention it because they'll see the roles if they query `pg_roles`. I don't feel strongly about it, though.

docs/cloud-tutorials/best-practices.md
maxdml: Yup, to revisit

docs/cloud-tutorials/best-practices.md
```diff
@@ -0,0 +1,91 @@
+---
+sidebar_position: 8
+title: Best practices
+description: DBOS Cloud Tips
+---
+
+A list of tips for using DBOS Cloud effectively.
+
+## Managing applications
+* Only one request at a time can manipulate an application and you will get an `application is busy` error if you attempt multiple concurrent modifications.
```
qianl15, line 10: We can provide a solution here, for example "If you see an `application is busy` error, try to wait for a couple of minutes before retrying".

docs/cloud-tutorials/best-practices.md
```diff
@@ -0,0 +1,91 @@
...
+* Application names are unique within an [organization](#users-and-organizations).
+* Your application is allocated 1GB of disk space, 1CPU and 512MB of RAM.
+
+### Deployment
+* An application must be registered before being deployed. DBOS Cloud CLI handles registration automatically when you deploy an application for the first time.
```
qianl15, line 15: We're removing app registration completely from our docs (the other PR). So we don't need this now.

docs/cloud-tutorials/application-management.md
```diff
@@ -39,6 +39,10 @@ When you deploy an application, DBOS Cloud first runs [`npx dbos migrate`](../ap
 It then starts your application.
 After your application is deployed, DBOS Cloud hosts it at this URL, which is also printed by the deploy command:
 
+:::info
+Each application is allocated 1GB of disk space.
```
apoliakov, line 43: I'm seeing 500MB here: 
https://github.com/dbos-inc/dbos-cloud/blob/733e24f95ec258e06f12151428e42f3fa952c03b/services/applications/deploy.go#L22

Is the code lying?

docs/cloud-tutorials/best-practices.md
```diff
@@ -0,0 +1,91 @@
...
+* All application archives -- including previous versions -- will be deleted from DBOS Cloud's artifact store.
+* Applications are assigned a [dedicated database role during deployment](#database-roles). Deleting an application will delete this role.
+* You can elect to drop the application database alongside deleting an application. This will also delete all the [database roles DBOS Cloud manages on behalf of the application](#database-roles).
+
+### Time travel
+Time travel can be enabled when deploying an application for the first time. Existing applications will need to be deleted and re-deployed to enable time travel.
+See more information in the [time travel](#time-travel) section.
+
+### Swapping database instance
+While an application cannot change database once registered, you can change the database instance (server) hosting the application database. When doing so, you can also decide to redeploy a previous version of the application. This is how DBOS Cloud support [point in time recovery](database-management#database-recovery) today.
```
apoliakov, line 36: It's hard for me to understand what's happening here. You are telling me:
1. I cannot change my database (do you mean logical db name?)
2. But I can change my database instance (server)

So what does that mean? Does it mean my data moves from one server to another? What happens to the database on the old server and how does the database on the new server come to be?

Nit: add an "s" to this is how DBOS Cloud support**s** point in time recovery 
...and I don't get the connection. Maybe if you tell me a bit more about how switching database servers works, it will become more apparent how it relates to recovery.

docs/cloud-tutorials/application-management.md
chuck-dbos: I'm still hung up on the idea of "disk space".  You shouldn't be getting "disk space".  Is it that the application image should be less than 1GB (or 500MB)?

docs/cloud-tutorials/best-practices.md
```diff
@@ -0,0 +1,91 @@
...
+### System databases
+* DBOS Transact relies on a _system database_ to store metadata about the application. This database name is `<app_db_name>_dbos_sys` and cannot be configured when running on DBOS Cloud.
+
+## Database instances
+
+### Provisioning instances
+* Provisioning instances is an asynchronous process. You can check the status of the instance with `npx dbos-cloud db status <database-instance-name>`.
+
+### Linked databases (BYODB)
+* DBOS Cloud will not configure `dbos_role` for linked database.
```
apoliakov, line 73: I think this is the expected flow, right?
https://docs.dbos.dev/cloud-tutorials/byod-management#linking-your-database-to-dbos-cloud

I'm not sure we need to say anything about BYODB on this page? Do we?

docs/cloud-tutorials/best-practices.md
```diff
@@ -0,0 +1,91 @@
...
+### Schema migrations
+* You can instruct DBOS Cloud to run [migration commands](../api-reference/configuration#database) during application deploy. These are handled by `dbosadmin`. You can use them to install extensions, create tables, etc.
+* For more complex operations, you can use the DBOS Cloud CLI to connect to your database and run migrations manually.
+
+### System databases
+* DBOS Transact relies on a _system database_ to store metadata about the application. This database name is `<app_db_name>_dbos_sys` and cannot be configured when running on DBOS Cloud.
+
+## Database instances
+
+### Provisioning instances
```
apoliakov, line 69:
```suggestion
### Provisioning database instances
```

docs/cloud-tutorials/best-practices.md
```diff
@@ -0,0 +1,91 @@
...
+
+:::tip
+Application databases are databases within a Postgres database instance (server).
+:::
+
+* Application databases cannot be shared across applications.
+* An application cannot change its database once deployed. You can delete the application and re-deploy with a new database name.
+
+### Database roles
+* DBOS Cloud will create the following roles for each application:
```
apoliakov, line 48: I don't know about this section. We don't list the privileges of dbosadmin but we list the privileges of `dbos_role` in great detail. That seems inconsistent. That aside, why are we listing those things to begin with? This does not seem like a "best practice" but more of a "explanation how the roles work" Fair and maybe useful but what do we want the user to do about it? How should the user use these roles and for what?

docs/cloud-tutorials/best-practices.md
```diff
@@ -0,0 +1,91 @@
...
+* DBOS Cloud will not configure `dbos_role` for linked database.
+* DBOS Cloud will not provide a default SSL certificate for linked database [FIXME: what is the expected flow here?].
+
+## Users and organizations
+
+* After logging in with DBOS Cloud, a `.dbos` folder is created. It hosts your current session information, including your credentials.
+
+### Registering a new user
+* New users need to be signed up with DBOS Cloud (through auth0) before registering a Cloud account.
+* New users will be assigned their eponymous organization by default. See [organizations](account-management#organization-management) for more information about organizations.
```
apoliakov, line 82: Similar comment. This is not a "Best practice" but "an explanation for how organizations work." Are there specific things the user should do (or not do?). If not, do we need to burden them with this extra detail?

docs/cloud-tutorials/best-practices.md
maxdml: it is hard to provide such a time frame because it dpends on what action is currently being done on the app. Deploys are 30seconds. Deletes a few seconds.

docs/cloud-tutorials/best-practices.md
maxdml: I removed it but the API does not work like that and I expect the question to be raised again in the future.

docs/cloud-tutorials/application-management.md
maxdml: > I'm seeing 500MB here: https://github.com/dbos-inc/dbos-cloud/blob/733e24f95ec258e06f12151428e42f3fa952c03b/services/applications/deploy.go#L22
> 
> Is the code lying?

No, you are right, but during the build step, we allow 1GB of disk to accommodate for dev dependencies. So a user could see >500MB from their app. But the deploy will fail if the archive size is more than 500MB. I will update now.*


> I'm still hung up on the idea of "disk space". You shouldn't be getting "disk space". Is it that the application image should be less than 1GB (or 500MB)?

I'll adjust based on Alex comment and say that `Each application can be up to 500MB in size`

maxdml: Also clarified this section

docs/cloud-tutorials/best-practices.md
maxdml: > (do you mean logical db name?)

Yes. "Logical DB name" is a term we came up with internally, but these are usually referred to as "databases"

> So what does that mean?

It is like a pointer change. The application is re-deployed with the new database as a "target". Only side effect is that PENDING workflows will be deleted -- because we do not migrate state from the previous DB instance to the new DB instance. I'll clarify.

docs/cloud-tutorials/application-management.md
```diff
@@ -11,25 +11,32 @@ In this guide, you'll learn how to manage applications in DBOS Cloud.
...
+```shell
 npx dbos-cloud app deploy
 ```
 
 Your application is deployed using the name in its `package.json`.
-Application names should be between 3 and 30 characters and must contain only lowercase letters and numbers, dashes (`-`), and underscores (`_`).
+Application names should be between 3 and 30 characters and must contain only lowercase letters and numbers, dashes (`-`), and underscores (`_`). Application names are unique within an [organization](#users-and-organizations).
 
-When you deploy an application for the first time, the command prompts you to choose which [database instance](../cloud-tutorials/database-management.md) to connect your app to, or to provision one if you have none.
+When you deploy an application for the first time, the command prompts you to choose which [database instance](../cloud-tutorials/database-management.md) to connect your app to, or to provision one if you have none. Note this database instance (server) is different from the database name your application use (the `app_db_name` field in [`dbos-config.yaml`](../api-reference/configuration#database)).
```
lanice, line 21:
```suggestion
When you deploy an application for the first time, the command prompts you to choose which [database instance](../cloud-tutorials/database-management.md) to connect your app to, or to provision one if you have none. Note this database instance (server) is different from the database name your application uses (the `app_db_name` field in [`dbos-config.yaml`](../api-reference/configuration#database)).
```

docs/cloud-tutorials/application-management.md
```diff
@@ -11,25 +11,32 @@ In this guide, you'll learn how to manage applications in DBOS Cloud.
...
 
 :::tip
-You can specify a particular database instance through the `-d <database-instance-name>`.
-To enable time travel for your application, specify `--enable-timetravel`.
+* You can specify a particular database instance through the `-d <database-instance-name>`.
+* During the first deployment, you can enable time travel for your application with `--enable-timetravel`. You can delete and re-deploy an existing application to enable time travel.
```
lanice, line 25: Just for my understanding: If I do not enable timetravel at first, it is not enough to just re-depoy with `--enable-timetravel`? I have to delete the application and re-deploy?

docs/cloud-tutorials/application-management.md
```diff
@@ -40,7 +47,9 @@ If your account belongs to an [organization](./account-management.md#organizatio
 If you edit your application or schema, run `npx dbos-cloud app deploy` again to apply the latest migrations and upgrade to the latest version.
 
 :::info
-You don't have to worry about changing database server connection parameters like `hostname` or `password` in [`dbos-config.yaml`](../api-reference/configuration.md) to deploy an application to the cloud&#8212;DBOS automatically applies the connection information of your cloud database instance.
+* You don't have to worry about changing database server connection parameters like `hostname` or `password` in [`dbos-config.yaml`](../api-reference/configuration.md) to deploy an application to the cloud&#8212;DBOS automatically applies the connection information of your cloud database instance.
+* You cannot change the database name of a deployed application. You can delete and re-deploy the application.
+* Databases cannot be shared across applications.
```
lanice, line 52: This **might** need clarification/differentiation between database instance and app db name, similar to what you did above.

docs/cloud-tutorials/application-management.md
```diff
@@ -66,18 +77,35 @@ npx dbos-cloud app versions <app-name>
...
 You can redeploy a previous version of your application by passing `--previous-version <version-id>` to the [`app deploy`](../api-reference/cloud-cli.md#npx-dbos-cloud-app-deploy) command.
 
-```
+```shell
 npx dbos-cloud app deploy --previous-version <version-id>
 ```
 
 However, this will fail if the previous and current versions have different database schemas.
+
+Also, note DBOS Cloud will keep running Transact workflows from previous versions until they complete. This means you need to be mindful of non-backward compatible database schema migrations, as they could break running workflows.
```
lanice, line 86: Do we have any safeguards/warnings in place for that? Or would it just break a running workflow from an old version if I deploy a non-backward compatible schema migration while an old workflow is still running?

docs/cloud-tutorials/database-management.md
```diff
@@ -4,37 +4,43 @@ title: Cloud Database Management
...
-Before you can deploy an application to DBOS Cloud, you must provision a Postgres database instance for it.
+Before you can deploy an application to DBOS Cloud, you must provision a Postgres database instance (server) for it.
 You must choose a database instance name, username and password.
-Both the database instance name and username must be 3 to 16 characters long and contain only lowercase letters, numbers and underscores.
-The database password must contain 8 or more characters.
+
+:::info
+* Both the database instance name and username must be 3 to 16 characters long and contain only lowercase letters, numbers and underscores.
+* The username must start with a letter.
+* The usernames `dbosadmin`, `dbos`, `postgres` and `admin` are reserved and cannot be used.
```
lanice, line 15: Good to know :sweat_smile: This is currently not handled in the console I think.

docs/cloud-tutorials/application-management.md
```diff
@@ -55,6 +64,8 @@ DBOS provides many tools to monitor and debug applications:
 
 - To retrieve the status of a particular application, run [`npx dbos-cloud app status <app-name>`](../api-reference/cloud-cli.md#npx-dbos-cloud-app-status). To retrieve the statuses of all applications, run [`npx dbos-cloud app list`](../api-reference/cloud-cli.md#npx-dbos-cloud-app-list).
 
+Applications are run in microVMs sized with 1vCPU and 512MB of RAM. Note that files on disk are ephemeral and not persisted between deployments.
```
kraftp, line 67:
```suggestion
Applications are run in [Firecracker microVMs](https://firecracker-microvm.github.io/) with 1vCPU and 512MB of RAM. Note that files on disk are ephemeral and not persisted between deployments.
```

docs/cloud-tutorials/application-management.md
kraftp: No safeguards at the moment. To be fair, this problem isn't unique to Transact--if you do rolling deployments of an app to any service, you'll hit the same problem with non-backwards-compatible migrations (which is why many shops won't do such migrations at all, they're too risky).

kraftp, line 86:
```suggestion
Also, note DBOS Cloud will keep running Transact workflows from previous versions until they complete. This means you need to be mindful of non-backward-compatible database schema migrations, as they could break running workflows.
```

docs/cloud-tutorials/application-management.md
```diff
@@ -66,18 +77,35 @@ npx dbos-cloud app versions <app-name>
...
 ### Deleting Applications
 
 To delete an application, run:
 
-```
+```shell
 npx dbos-cloud app delete <app-name>
 ```
+
+The application will be fully deleted once all its running workflows complete.
```
kraftp, line 99: It's deleted immediately, workflows or no

kraftp (commented):
I really like what this PR is now doing!  Just a bit of cleanup to go.

docs/cloud-tutorials/application-management.md
maxdml: yup you have to delete + redeploy

docs/cloud-tutorials/database-management.md
maxdml: Let's fix it ;-)

docs/cloud-tutorials/application-management.md
maxdml: What I mean is that the application will be in state "DELETED", and a record still present & listed in `dbos-cloud app list`, until all workflows are completed

docs/cloud-tutorials/application-management.md
```diff
@@ -11,25 +11,32 @@ In this guide, you'll learn how to manage applications in DBOS Cloud.
...
+```shell
 npx dbos-cloud app deploy
 ```
 
 Your application is deployed using the name in its `package.json`.
-Application names should be between 3 and 30 characters and must contain only lowercase letters and numbers, dashes (`-`), and underscores (`_`).
+Application names should be between 3 and 30 characters and must contain only lowercase letters and numbers, dashes (`-`), and underscores (`_`). Application names are unique within an [organization](account-management#organization-management).
 
-When you deploy an application for the first time, the command prompts you to choose which [database instance](../cloud-tutorials/database-management.md) to connect your app to, or to provision one if you have none.
+When you deploy an application for the first time, the command prompts you to choose which [database instance](../cloud-tutorials/database-management.md) to connect your app to, or to provision one if you have none. Note this database instance (server) is different from the database name your application uses (the `app_db_name` field in [`dbos-config.yaml`](../api-reference/configuration#database)).
```
kraftp, line 21: Note that multiple applications can connect to the same database server, but must use separate databases within that server (app_db_name...)

docs/cloud-tutorials/application-management.md
```diff
@@ -40,7 +47,9 @@ If your account belongs to an [organization](./account-management.md#organizatio
 If you edit your application or schema, run `npx dbos-cloud app deploy` again to apply the latest migrations and upgrade to the latest version.
 
 :::info
-You don't have to worry about changing database server connection parameters like `hostname` or `password` in [`dbos-config.yaml`](../api-reference/configuration.md) to deploy an application to the cloud&#8212;DBOS automatically applies the connection information of your cloud database instance.
+* You don't have to worry about changing database server connection parameters like `hostname` or `password` in [`dbos-config.yaml`](../api-reference/configuration.md) to deploy an application to the cloud&#8212;DBOS automatically applies the connection information of your cloud database instance.
+* You cannot change the database name (`app_db_name`) of a deployed application. You can delete and re-deploy the application.
```
kraftp, line 51:
```suggestion
* You cannot change the application database (`app_db_name`) of a deployed application. You must delete and re-deploy the application.
```

docs/cloud-tutorials/application-management.md
```diff
@@ -40,7 +47,9 @@ If your account belongs to an [organization](./account-management.md#organizatio
...
+* Databases (`app_db_name`) cannot be shared across applications.
```
kraftp, line 52:
```suggestion
* While database instances can be shared across applications, application databases (`app_db_name`) cannot.
```

docs/cloud-tutorials/application-management.md
kraftp: No, it won't, we changed that behavior (or if we didn't, we should), otherwise long-running workflows would block deletion forever

docs/cloud-tutorials/application-management.md
```diff
@@ -66,18 +77,35 @@ npx dbos-cloud app versions <app-name>
...
+
+The application will be fully deleted once all its running workflows complete.
+
+You can also delete the application database with the `--dropdb` argument:
+
+```shell
+npx dbos-cloud app delete <app-name> --dropdb
+```
+
+This will delete the database and all associated DBOS roles.
```
kraftp, line 107:
```suggestion
```

kraftp (approved):

PR #191, opened 2024-09-04
DBOS Class Docs

kraftp:
- Document main `dbos.DBOS` class
- Update code examples
- "AI-Powered Slackbot"

### DIFF
--- a/docs/python/reference/dbos-class.md
+++ b/docs/python/reference/dbos-class.md
@@ -25,7 +25,7 @@ DBOS(
 ### launch
 
 ```python
-dbos.launch()
+DBOS.launch()
 ```
 
 Launch DBOS, initializing database connections and starting scheduled workflows.
@@ -47,7 +47,7 @@ def run_every_minute(scheduled_time: datetime, actual_time: datetime):
 
 # In your main function, launch DBOS
 if __name__ == "__main__":
-    dbos.launch()
+    DBOS.launch()
 ```
 
 ### destroy
--- a/docs/python/reference/dbos-class.md
+++ b/docs/python/reference/dbos-class.md
@@ -37,7 +37,7 @@ If a FastAPI app is passed into the `DBOS` constructor, `launch` is called autom
 from dbos import DBOS
 
 # Initialize the DBOS object
-dbos = DBOS()
+DBOS()
 
 # Define a scheduled workflow
 @DBOS.scheduled("* * * * *")
--- a/docs/python/reference/dbos-class.md
+++ b/docs/python/reference/dbos-class.md
@@ -3,7 +3,7 @@ sidebar_position: 1
 title: DBOS Class
 ---
 
-The DBOS class is a singleton&mdash;you may only instantiate it once in a program's lifetime.
+The DBOS class is a singleton&mdash;you must instantiate it (by calling its constructor) exactly once in a program's lifetime.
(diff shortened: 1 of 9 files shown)


docs/python/reference/dbos-class.md
```diff
@@ -0,0 +1,61 @@
+---
+sidebar_position: 1
+title: DBOS Class
+---
+
+The DBOS class is a singleton&mdash;you may only instantiate it once in a program's lifetime.
```
qianl15, line 6: "initiate" is a bit vague. We can say "you may only call its constructor exactly once in a program's lifetyle"

qianl15: If I understand it correctly, you always have to call `DBOS()` even if you're explicitly using it?

docs/python/reference/dbos-class.md
```diff
@@ -0,0 +1,61 @@
...
+```python
+DBOS(
+    *,
+    fastapi: Optional[FastAPI] = None,
+    config: Optional[ConfigFile] = None,
+)
+```
+
+**Parameters:**
+- `fastapi`: If your application is using FastAPI, the `FastAPI` object. If this is passed in, DBOS automatically calls `launch` when FastAPI is fully initialized. DBOS also adds to all routes a middleware that enables [tracing](../tutorials/logging-and-tracing.md#tracing) through FastAPI HTTP endpoints.
```
qianl15, line 21: Provide a pointer to "launch". It's unclear which launch it calls, FastAPI or itself.

docs/python/reference/dbos-class.md
```diff
@@ -0,0 +1,61 @@
...
+- `config`: A configuration object, which if passed in is used in place of `dbos-config.yaml`. We recommend using this for testing only.
```
qianl15, line 22: Explicitly say: we use dbos-config.yaml file by default.

docs/python/reference/dbos-class.md
```diff
@@ -0,0 +1,61 @@
...
+
+# Define a scheduled workflow
+@DBOS.scheduled("* * * * *")
+@DBOS.workflow()
+def run_every_minute(scheduled_time: datetime, actual_time: datetime):
+    DBOS.logger.info("This is a scheduled workflow!")
+
+# In your main function, launch DBOS
+if __name__ == "__main__":
+    dbos.launch()
```
qianl15, line 50: It does look weird that we use `dbos.launch()` but `DBOS.destroy()`. This small dbos vs big DBOS still look weird. Why can't we just use `DBOS.launch()`?

docs/python/reference/dbos-class.md
kraftp: Fixed

docs/python/reference/dbos-class.md
```diff
@@ -0,0 +1,61 @@
...
+Launch DBOS, initializing database connections and starting scheduled workflows.
+Should be called in your main function.
+If a FastAPI app is passed into the `DBOS` constructor, `launch` is called automatically during FastAPI setup.
+
+**Example:**
+```python
+from dbos import DBOS
+
+# Initialize the DBOS object
+dbos = DBOS()
```
qianl15: No small dbos

qianl15 (approved):

PR #199, opened 2024-09-07
quickstart fixes

devhawk:


docs/quickstart.md
```diff
@@ -37,6 +37,11 @@ python3 -m venv .venv
 source .venv/bin/activate
 ```
 
+:::info
+Windows users will need to activate the virtual environment with `.venv\Scripts\activate.ps1` or `.venv\Scripts\activate.bat`.
+See the [official Python documentation](https://docs.python.org/3/library/venv.html#how-venvs-work) for more information.
+:::
+
```
chuck-dbos, line 44: Should this be a tab?  `source` doesn't work in cmd...

docs/quickstart.md
```diff
@@ -37,6 +37,11 @@ python3 -m venv .venv
 source .venv/bin/activate
 ```
 
+:::info
```
kraftp, line 40: Can we instead do this with tabs? Like in some of the cross-platform examples below?  Looks much cleaner in my opinion.

chuck-dbos (approved):

kraftp (commented):
I'll merge this immediately because we just had a user get tripped up by it, then we'll edit later

PR #206, opened 2024-09-09
Better Python Guide

kraftp:


docs/python/programming-guide.md
```diff
@@ -7,321 +7,177 @@ pagination_prev: quickstart
...
+def insert_greeting(name: str) -> str:
+    query = dbos_hello.insert().values(name=name)
+    DBOS.sql_session.execute(query)
     DBOS.logger.info(f">>> STEP 2: Greeting to {name} recorded in the database!")
 
-
+@app.get("/greeting/{name}")
 @DBOS.workflow()
-def greeting_workflow(name: str, note: str):
+def greeting_endpoint(name: str):
```
apoliakov, line 142: Reminder just in case: there's a problem here that caused us to use `start_workflow`. On Chrome, the browser would try to reconnect after ctrl+c which resulted in two workflows and confusing output.

qianl15 (approved):

PR #224, opened 2024-09-22
Queue

kraftp:


docs/python/reference/task_queues.md
```diff
@@ -17,9 +17,9 @@ Queue(
 
 **Parameters:**
 - `name`: The name of the queue. Must be unique among all your queues.
-- `concurrency`: The maximum number of tasks from this queue that may run concurrently.
+- `concurrency`: The maximum number of operations from this queue that may run concurrently.
```
chuck-dbos, line 20: workflows?

chuck-dbos: You get a workflow handle back, right?

kraftp: They can also be individual steps or transactions (technically wrapped in workflows).

chuck-dbos: Functions, maybe, because they are that.  Or is there a collective nown we already use for workflows/transactions/steps... otherwise I'd hate to define a new (highly overloaded) term like jobs/tasks/operations.

Otherwise, workflows is technically correct, and you get workflow behavior (like comm retries), and a workflow handle, and all that... so while it may surprise people at first, it's correct to say workflow queue.

kraftp: Okay switched them all to functions!

chuck-dbos (approved):

PR #225, opened 2024-09-23
Auth tutorial for Python

chuck-dbos:
For #226 

Some examples of FastAPI middleware and how they work with DBOS role decorators


docs/python/tutorials/authentication-authorization.md
```diff
@@ -0,0 +1,127 @@
...
+Authentication information may arrive in various ways, depending on the approach, protocol, and framework used.  It is common to use some sort of "middleware" that sits between the request handler and the application code that processes the authentication information in each request.
+
+A simple middleware for FastAPI may pass authentication information to DBOS in a manner similar to this:
+```python
+    @app.middleware("http")
+    async def authMiddleware(
+        request: Request, call_next: Callable[[Request], Awaitable[Response]]
+    ) -> Response:
+        with DBOSContextEnsure():
+            prev_user = DBOS.authenticated_user
```
kraftp, line 21: What makes the `prev_user` switchout necessary? Can we eliminate the need for it?  It seems overcomplicated.

docs/python/tutorials/authentication-authorization.md
```diff
@@ -0,0 +1,127 @@
...
+---
+
+DBOS Python supports modular, declarative security.  This is a cooperative effort between the request framework (such as FastAPI) and DBOS Transact; the server framework performs authentication and forwards the authenticated user and roles on to DBOS functions, which then check for authorization.  Authentication information is forwarded via the [DBOS context](../reference/contexts.md#set_authentication).
+
+The following tutorial shows use of FastAPI middleware to collect and authenticate user information, and shows how to set function- and class-level authorization in DBOS.
+
+## Authentication Middleware
+Authentication information may arrive in various ways, depending on the approach, protocol, and framework used.  It is common to use some sort of "middleware" that sits between the request handler and the application code that processes the authentication information in each request.
+
+A simple middleware for FastAPI may pass authentication information to DBOS in a manner similar to this:
```
kraftp, line 14:
```suggestion
A simple middleware for FastAPI may pass authentication information to DBOS in a manner similar to this. You can construct analogous middleware in Flask or other web frameworks.
```

docs/python/tutorials/authentication-authorization.md
chuck-dbos: There are risks with leaving behind permissions.  We could do it a different way, but we should clear it out when the scope ends so you don't have to wonder what would happen if they are left in a context... where does that context go next, could that be abused as a security hole, etc....

docs/python/tutorials/authentication-authorization.md
chuck-dbos: > 

Should we just try it and paste a flask example in below ?

docs/python/tutorials/authentication-authorization.md
chuck-dbos: We could let DBOSContextEnsure or similar do it.

docs/python/tutorials/authentication-authorization.md
kraftp: Yeah we could tab FastAPI and Flask

docs/python/tutorials/authentication-authorization.md
kraftp: Contexts aren't shared across requests, are they?  If they are we should fix it.

chuck-dbos: This is the kind of code that ensures that.  We don't know if ContextVar, thread, activation, whatever you want to call it is reused across requests... the server code in FastAPI knows that.  Our middleware will balance it out so nothing gets reused when it shouldn't.

kraftp: This shouldn't be up to the user, we should do the right thing automatically if FastAPI doesn't (do they?)

chuck-dbos: The solution is to give the user / middleware author a better utility, not to count on whatever else is in the front end / middleware stack.

kraftp (approved):

PR #232, opened 2024-09-27
Rate Limiting

kraftp:


docs/python/tutorials/queue-tutorial.md
```diff
@@ -67,10 +67,22 @@ from dbos import Queue
...
 ```
 
-You may want to specify a maximum concurrency if functions in your queue submit work to an external process or call a rate-limited API.
-The concurrency limit guarantees that even if many functions are submitted at once, they won't overwhelm the worker process or trigger the API's rate limit.
+You may want to specify a maximum concurrency if functions in your queue submit work to an external process with limited resources.
+The concurrency limit guarantees that even if many functions are submitted at once, they won't overwhelm the process.
+
+### Rate Limiting
+
+You can set _rate limits_ for a queue, limiting the number of functions that it can start in a given period.
```
qianl15, line 75: Make sure to mention that we don't drop functions

qianl15 (approved):

PR #247, opened 2024-10-16
dbos-cloud app update documentation

maxdml:
- Update Cloud CLI reference
- Update application management tutorial

Screenshot 2024-10-15 at 20 43 03
Screenshot 2024-10-15 at 20 42 43

Screenshot 2024-10-16 at 10 12 58


kraftp (commented):
We also need to update the docs where it says app VM sizes are always 512MB to now say that's configurable.

maxdml:
> We also need to update the docs where it says app VM sizes are always 512MB to now say that's configurable.

Done

kraftp (approved):

PR #249, opened 2024-10-17
Visually emphasis the requirement to have a local Postgres instance in the Python quickstart

maxdml:
- Visually emphasis the requirement to have a local Postgres instance, in the Python quickstart.
- Also specify that PGPASSWORD must be exporter in the environment (another assumption the guide makes) 

I am not updating the TS quickstart because it likely requires a larger update.

### DIFF
--- a/docs/python/programming-guide.md
+++ b/docs/python/programming-guide.md
@@ -10,8 +10,8 @@ First, without using DBOS, we'll build an app that records greetings to two diff
 Then, we'll add DBOS durable execution to the app in **just four lines of code**.
 Thanks to durable execution, the app will always write to both systems consistently, even if it is interrupted or restarted at any point.
 
-This guide assumes you have a Postgres database running locally.
-If not, see the [quickstart](../quickstart.md) for instructions on how to set it up.
+_**This guide assumes you have a Postgres database running locally and PGPASSWORD is exported in your environment.**_
+_**If not, see the [quickstart](../quickstart.md) for instructions on how to set it up.**_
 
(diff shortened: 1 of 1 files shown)


docs/python/programming-guide.md
```diff
@@ -10,8 +10,8 @@ First, without using DBOS, we'll build an app that records greetings to two diff
 Then, we'll add DBOS durable execution to the app in **just four lines of code**.
 Thanks to durable execution, the app will always write to both systems consistently, even if it is interrupted or restarted at any point.
 
-This guide assumes you have a Postgres database running locally.
-If not, see the [quickstart](../quickstart.md) for instructions on how to set it up.
+_**This guide assumes you have a Postgres database running locally and PGPASSWORD is exported in your environment.**_
```
kraftp, line 13: I would take out the part about PGPASSWORD-- if they're doing the new local-cloud-database thing then they don't need PGPASSWORD.

maxdml: Maybe we should reword as follows, to be inclusive:

`This guide assumes you have a Postgres database running locally. You might need to export PGPASSWORD in your environment. See the [quickstart](../quickstart.md) for instructions on how to set up Postgres.`

maxdml: Updated. Though we might need to revisit in light of upcoming PG requirements changes?

kraftp:
Guide was changed around in https://github.com/dbos-inc/dbos-docs/pull/252 to no longer need this

PR #250, opened 2024-10-17
Document Detective

kraftp:


docs/python/examples/document-detective.md
```diff
@@ -0,0 +1,275 @@
...
+```
+
+## Building a Durable Data Ingestion Pipeline
+
+Now, let's write the document ingestion pipeline.
+Because ingesting and indexing documents may take a long time, we need to build a pipeline that's both _concurrent_ and _reliable_.
+It needs to process multiple documents at once and it needs to be resilient to failures, so if the application is interrupted or restarted, or encounters an error, it can recover from where it left off instead of restarting from the beginning or losing some documents entirely.
+
+We'll build a concurrent, reliable data ingestion pipeline using DBOS [queues](../tutorials/queue-tutorial.md) and [durable execution](../tutorials/workflow-tutorial.md).
+This workflow takes in a batch of document URLs and enqueues them for ingestion.
```
qianl15, line 73: Mention that there will be multiple threads processing this queue.

PR #256, opened 2024-11-01
document stock tracker app

maxdml:
Screenshot 2024-12-09 at 12 12 54

### DIFF
--- a/docs/python/examples/stock-tracker.md
+++ b/docs/python/examples/stock-tracker.md
@@ -64,7 +64,7 @@ def fetch_stock_price(symbol):
 
 ## Sending SMS alerts
 
-We'll use the Twilio API to send SMS alerts and write a DBOS [step](../tutorials/step-tutorial) to make it durable. You can sign up for a free Twilio account at https://www.twilio.com/try-twilio. We'll need a few environment variables to store our Twilio account SID, auth token and phone number.
+We'll use the Twilio API to send SMS alerts and write a DBOS [step](../tutorials/step-tutorial) to make it durable. You can sign up for a free Twilio account at https://www.twilio.com/try-twilio. We'll need a few environment variables to store our Twilio account SID, auth token and phone number. The alerts themselves are configured on the app's streamlit frontend.
 
 ```python
 twilio_account_sid = os.environ.get('TWILIO_ACCOUNT_SID')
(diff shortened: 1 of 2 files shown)


docs/python/examples/stock-tracker.md
```diff
@@ -0,0 +1,108 @@
+---
+displayed_sidebar: examplesSidebar
+sidebar_position: 10
+title: Stock Tracker
+---
+
+In this example, we use DBOS to build and deploy an app that tracks stock prices and send SMS alerts when the price of a stock crosses a certain threshold. We certainly don't want to miss a good opportunity to trade, so it's important this workflow is punctual and reliable!
+
+You can see the application live [here](https://max-stock-prices.cloud.dbos.dev/). All source code is [available on GitHub](https://github.com/dbos-inc/dbos-demo-apps/tree/main/python/stock-prices).
```
kraftp, line 9: Do the alerts actually work?  Who are they sent to?

docs/python/examples/stock-tracker.md
```diff
@@ -0,0 +1,108 @@
...
+    DBOS.logger.info(f"Stock price for {symbol} is {data['Close'].iloc[0]}")
+    return data['Close'].iloc[0]
+```
+
+## Sending SMS alerts
+
+We'll use the Twilio API to send SMS alerts and write a DBOS [step](../tutorials/step-tutorial) to make it durable. You can sign up for a free Twilio account at https://www.twilio.com/try-twilio. We'll need a few environment variables to store our Twilio account SID, auth token and phone number.
+
+```python
+twilio_account_sid = os.environ.get('TWILIO_ACCOUNT_SID')
```
kraftp, line 70: Where do the alerts come from?

docs/python/examples/stock-tracker.md
maxdml: The SMS are sent to Twilio. You can see them in your twilio account.

The only issue is that, with a free account, they don't get actually sent to the recipient. If you upgrade to a paying account, then they'll go through.

docs/python/examples/stock-tracker.md
maxdml: The DBOS workflow will call the Twilio API. Twilio is responsible for delivering the SMS.

kraftp (approved):
This needs a tile on the examples page, but otherwise looks good!

PR #268, opened 2024-12-04
Initial doc for TSV2

chuck-dbos:
Initial reference docs for TS V2 API

Fix a couple of small issues w/ older docs


docs/typescript/reference/decorators.md
qianl15, line 1: Do we want to put a warning at the top saying this page is for `v1`? Then redirect people to the newer version.

qianl15: Maybe we should move the old reference docs like DBOS Decorators, DBOS Contexts, and Testing Runtime into a separate folder?

docs/typescript/reference/dbos.md
```diff
@@ -0,0 +1,953 @@
...
+sidebar_position: 4
+title: DBOS Class
+description: API reference for DBOS
+---
+
+# DBOS Class
+
+The `DBOS` class is the heart of the DBOS Transact programming API, providing decorators to identify key application functions, and accessors for context-dependent state.
+
+Note that this is the second major version of the API.  The first version of the API was based on passing [contexts](./contexts.md) and used a separate set of [decorators](./decorators.md).  It is fine to use both approaches in the same DBOS application, however the `DBOS` class is simpler and is therefore recommended.
```
qianl15, line 11: Put this in a highlighted box using `:::info`

docs/typescript/reference/decorators.md
kraftp: Agreed. This is a fresh start in my opinion, although we maintain compatibility with the old version. I would totally remove all references to the old version then split up this huge file. That way we get a much cleaner structure as opposed to 6 old pages and one gigantic new page.

docs/typescript/reference/dbos.md
qianl15, line 1: The python version uses the name `dbos-class.md`. I think we can make it consistent here.

docs/typescript/reference/dbos.md
```diff
@@ -0,0 +1,953 @@
+---
+sidebar_position: 4
+title: DBOS Class
+description: API reference for DBOS
+---
+
+# DBOS Class
+
+The `DBOS` class is the heart of the DBOS Transact programming API, providing decorators to identify key application functions, and accessors for context-dependent state.
```
qianl15, line 9: If I understand it correctly, the `DBOS` class contains only static method. So there's no way to instantiate a DBOS object like we do in Python. If so, we might want to clarify that this class only contains static methods.

docs/typescript/reference/dbos.md
```diff
@@ -0,0 +1,953 @@
...
+import { DBOS, ConfiguredInstance } from '@dbos-inc/dbos-sdk';
+
+class MyClass extends ConfiguredInstance
+{
+  constructor(name: string) {
+    super(name);
+  }
+
+  async initialize() {
+    // Do initialization work
```
qianl15, line 80: There's no `InitContext` required anymore?

docs/typescript/reference/dbos.md
```diff
@@ -0,0 +1,953 @@
...
+export abstract class ConfiguredInstance {
+  readonly name: string;
+  constructor(name: string) {
+    this.name = name;
+  }
+  abstract initialize(ctx: InitContext): Promise<void>;
+}
+```
+
+The `ConfiguredInstance` base class provides naming information, and also declares the `initialize` function which will be called after the DBOS runtime has started, but before workflow processing commences.  The argument to `initialize` is an [`InitContext`](../reference/contexts.md#initcontext), which provides access to configuration and other information.
```
qianl15, line 67: Since we're getting away from contexts, do we want to remove the need of `InitContext` here as well? We can directly obtain connections through the `DBOS` class.

docs/typescript/reference/dbos.md
```diff
@@ -0,0 +1,953 @@
...
+```
+
+When called, `workflow` functions are wrapped such that they will be [run to completion exactly once](../tutorials/idempotency-tutorial.md).
+
+#### Decorating Static Methods
+Workflow functions must be class members, and may be `static`.  To mark a `static` member function as a workflow, simply place the `@DBOS.workflow()` decorator above it.
+
+Example:
+```typescript
+import { DBOS, ConfiguredInstance, WorkflowQueue } from '@dbos-inc/dbos-sdk';
```
qianl15, line 30:
```suggestion
import { DBOS } from '@dbos-inc/dbos-sdk';
```
I'd remove unused imports. Otherwise, it could confuse users.

docs/typescript/reference/dbos.md
```diff
@@ -0,0 +1,953 @@
...
+# DBOS Class
+
+The `DBOS` class is the heart of the DBOS Transact programming API, providing decorators to identify key application functions, and accessors for context-dependent state.
+
+Note that this is the second major version of the API.  The first version of the API was based on passing [contexts](./contexts.md) and used a separate set of [decorators](./decorators.md).  It is fine to use both approaches in the same DBOS application, however the `DBOS` class is simpler and is therefore recommended.
+
+## Decorating Workflows, Transactions, and Steps
+DBOS provides reliability guarantees for programs that are [written as workflows comprised of steps, transactions, and child workflows](../programming-guide.md).  The workflow, transaction, and step functions are marked with decorators.
+
+### DBOS.workflow
```
qianl15, line 16: The styling is not consistent here. We can consistently use `@` in front of decorators, and make methods in `code blocks`. 
```suggestion
### `@DBOS.workflow`
```

docs/typescript/reference/dbos.md
```diff
@@ -0,0 +1,953 @@
...
+The `DBOS.withWorkflowQueue` function runs the provided `callback` function and returns its result.  Any workflows started within `callback` will be run on the queue named by `wfq`.
+
+Example:
+```typescript
+// Create a workflow queue for use below
+const wfq = new WorkflowQueue("example_queue", 10, {limitPerPeriod: 50, periodSec: 30});
+
+// Run a workflow on the queue, and await the result
+const qres = await DBOS.withWorkflowQueue(wfq.name, async ()=>{
+  return await Workflows.processWorkflow("value q");
```
qianl15, line 172: So does this mean that the callback function can be anything and the return value is not necessarily a handle? Can we make `withWorkflowQueue` the same as `startWorkflow`?

qianl15: In this case, the workflow is not even running on a queue. It's not rate limited or processed asynchronously.

docs/typescript/reference/dbos.md
chuck-dbos: We can, that would involve a change to initialize in the codebase before we change the doc.

chuck-dbos: It's on my list, was just lower priority, I got convinced people are not using init much.

docs/typescript/reference/dbos.md
chuck-dbos: Yes, I reflowed this a few times, that import was supposed to be with the example that covers all sections below.  Fixing.

docs/typescript/reference/dbos.md
chuck-dbos: There are quite a number of these that I found on review, will do all in one commit.

docs/typescript/reference/dbos.md
chuck-dbos: I have in my TODO notes that startWorkflow could take more arguments, as it does in Python, including the queue.  But doing it this way, you could run a whole list of things, and any workflows would be run on queues, so I thought it was a nice addition.

docs/typescript/reference/decorators.md
chuck-dbos: I was thinking about this as a phased process... Having the API and this material as a reference, publishing to npm.  Then having the examples and changing all the tutorial sections to do the new way as a follow-on.  In case we find any issues, etc.

docs/typescript/reference/dbos.md
chuck-dbos: It should be running on a queue.

https://github.com/dbos-inc/dbos-transact-ts/blob/35ae800e1427051f406999e8818b8c73356ba3cc/tests/contextfreeinst.test.ts#L222

If not, we can fix it.  It's a Promise, so it's processed asynchronously.

docs/typescript/reference/dbos.md
qianl15: Sounds good.

docs/typescript/reference/dbos-class.md
```diff
@@ -0,0 +1,957 @@
...
+```
+
+##### Assigning workflow IDs
+Specifying a workflow ID is useful if the intent is to start a workflow exactly once for a given circumstance.  Assignment of [workflow IDs](../tutorials/idempotency-tutorial.md) can be done with `DBOS.withNextWorkflowID`.
+
+```typescript
+DBOS.withNextWorkflowID<R>(wfid: string, callback: ()=>Promise<R>) : Promise<R>
+```
+
+The `DBOS.withNextWorkflowID` function runs the provided `callback` function and returns its result.  The first workflow started within `callback` will be assigned the ID specified by `wfid`.
```
qianl15, line 194: The callback function currently allows running multiple workflows with the same workflow ID(?) Can we have a clear error message to disallow this?

docs/typescript/reference/dbos-class.md
```diff
@@ -0,0 +1,957 @@
...
+// Assign an idempotency key to the workflow, run asynchronously
+const handle6 = await DBOS.withNextWorkflowID('<wfid_value>', async() => {
+  return await DBOS.startWorkflow(Workflows).processWorkflow("value6");
+});
+// ...
+const result6 = await handle6.getResult();
+```
+
+#### Workflow Recovery Configuration
+`@DBOS.workflow()` takes an optional `WorkflowConfig` object:
```
qianl15, line 211: I'll explain `WorkflowConfig` right after `@DBOS.workflow`

docs/typescript/reference/dbos-class.md
```diff
@@ -0,0 +1,957 @@
...
+const res = DBOS.invoke(MyClass).transactionFunction('arg');
+```
+
+## Sleeping
+
+Within a DBOS workflow, waiting or sleeping should not be done using standard system functions, as these will not be skipped on workflow replay.  Instead, the "durable sleep" functions below should be used.  The wakeup time will be stored in the database when the function is first called, and if the workflow is re-executed, it will not oversleep.
+
+```typescript
+DBOS.sleepms(durationMS: number): Promise<void>
+DBOS.sleep(durationSec: number): Promise<void>
```
qianl15, line 336: Is now a good time to change `DBOS.sleep` to use millisecond by default? It's a new interface anyways...

docs/typescript/reference/dbos-class.md
```diff
@@ -0,0 +1,957 @@
...
+  static async transactionFunction(ctx: TransactionContext, arg: string) {
+    //...
+  }
+}
+
+// Call with `DBOS.invoke`
+const res = DBOS.invoke(MyClass).transactionFunction('arg');
+```
+
+## Sleeping
```
qianl15, line 330:
```suggestion
## Durable Sleep
```

docs/typescript/reference/dbos-class.md
```diff
@@ -0,0 +1,957 @@
...
+
+`DBOS.send` and `DBOS.recv` allows the sending of messages to a specific [workflow](../tutorials/workflow-tutorial#workflow-identity).  Workflows may wait for the message to be received before proceeding.
+
+#### `DBOS.send`
+```typescript
+DBOS.send<T>(destinationID: string, message: T, topic?: string): Promise<void>
+```
+`DBOS.send()` to send a message to a workflow, identified by `destinationID`.  Messages can optionally be associated with a topic and are queued on the receiver per topic.  `DBOS.send` may be called from other workflows, or anywhere else, to pass information to a corresponding `DBOS.recv` call.
+
+For more information, see our [messages API tutorial](../tutorials/workflow-communication-tutorial#messages-api).
```
qianl15, line 356: This link is about using contexts from v1. Maybe remove this reference for now.

docs/typescript/reference/dbos-class.md
```diff
@@ -0,0 +1,957 @@
...
+
+#### Reliability Guarantees
+
+All events are persisted to the database, so once an event it set, it is guaranteed to always be retrievable.
+
+## Finding and Retrieving Workflows
+Information about workflows that are either running or have already been completed can be retrieved from the workflow ID.  The workflow history can also be searched by other criteria.
+
+### `DBOS.getWorkflowStatus`
+`DBOS.getWorkflowStatus` retrieves the status of a single workflow, given its workflow ID.  If
```
qianl15, line 404:
```suggestion
`DBOS.getWorkflowStatus` retrieves the status of a single workflow, given its workflow ID.
```

docs/typescript/reference/dbos-class.md
```diff
@@ -0,0 +1,957 @@
...
+```
+
+### `DBOS.retrieveWorkflow`
+Similar to [`DBOS.getWorkflowStatus`](#dbosgetworkflowstatus), `DBOS.retrieveWorkflow` retrieves a single workflow by its id, but returns a [`WorkflowHandle`](../reference/workflow-handles.md).  This handle provides status, but also allows other workflow interactions.
+
+```typescript
+DBOS.retrieveWorkflow(workflowID: string)
+```
+
+### Searching for workflows
```
qianl15, line 432: I think we can remove this subtitle and move `DBOS.getWorkflows` and `DBOS.getWorkflowQueue` to the upper level.

docs/typescript/reference/dbos-class.md
```diff
@@ -0,0 +1,957 @@
...
+* `DBOS.recoverPendingWorkflows`: Retrieves a list of workflows to recover, and starts them, returning their handles
+* `DBOS.executeWorkflowById`: Starts executing a workflow given its `workflowId`.  If `startNewWorkflow` is `true`, the workflow is cloned and assigned a new unique workflow ID.
+
+## Extensibility
+DBOS Transact allows libraries, such as those designed to receive external events, to initiate DBOS functions.
+
+### Decorators
+Such libraries may choose to implement method decorators.  If the method decorator is registering a function with DBOS Transact, it should call `DBOS.registerAndWrapContextFreeFunction` for functions that do not accept a context as the first parameter.
+
+```typescript
```
qianl15, line 922: Would benefit from having an example here to show how to define a custom decorator.

docs/typescript/reference/dbos-class.md
```diff
@@ -0,0 +1,957 @@
...
+DBOS.registerAndWrapContextFreeFunction<This, Args extends unknown[], Return>(
+    target: object,
+    propertyKey: string,
+    descriptor: TypedPropertyDescriptor<(this: This, ...args: Args) => Promise<Return>>,
+)
+```
+
+`DBOS.registerAndWrapContextFreeFunction` registers the function named `propertyKey` on the `target` class, creates a wrapper function, and updates the `descriptor` to use the wrapper function.
```
qianl15, line 930: The user-facing name could be shorter, like `DBOS.createDecorator` or so...

docs/typescript/reference/dbos-class.md
```diff
@@ -0,0 +1,957 @@
...
+DBOS functions that are decorated as `DBOS.workflow`, `DBOS.transaction`, or `DBOS.step`, can be run directly from event receiver dispatch loops.
+
+For more complex circumstances, such as invoking functions with a context argument, `DBOS.executor` provides access to the [`DBOSExecutorContext` interface](../reference/contexts.md#dbosexecutorcontext).
+
+#### Setting Authenticated User And Roles
+When calling DBOS functions, it may be desired to set context variables, such as the authenticated user, and the current tracing span.  The following functions can be used for these purposes:
+
+```typescript
+// Set the authenticated user, and roles allowed by the authentication system for the scope of the callback() function
+DBOS.withAuthedContext<R>(authedUser: string, authedRoles: string[], callback: () => Promise<R>): Promise<R>
```
qianl15, line 942: I think instead of having separate contexts, can we create a single context type with the auth, trace, and caller name fields? Then, it will also allow users to attach their custom fields into the context object.

docs/typescript/reference/dbos-class.md
chuck-dbos: I struggled with the flow.  It's not really commonly used information so I found the gap to be acceptable.  Maybe I could link it.

docs/typescript/reference/dbos-class.md
chuck-dbos: I was thinking we would rewrite this, and other things that we're referring to, in the next round - up for review early next week and merged soon thereafter.  A link to the old one is better than none, for now.

docs/typescript/reference/dbos-class.md
chuck-dbos: There are 4 functions in the section.  I am not sure if you meant pull them up 1 level or up 2 levels to the upper level.  Trying 1 level, which I think will work out.

docs/typescript/reference/dbos-class.md
chuck-dbos: This needs to be tried before we'd have that.  I think we deprioritized custom event receivers.

docs/typescript/reference/dbos-class.md
chuck-dbos: Could possibly be added.

chuck-dbos: Also see comment above about needing more real experience with it before the API and doc will be very good.

docs/typescript/reference/dbos.md
chuck-dbos: Parameters added to startWorkflow.

docs/typescript/reference/dbos-class.md
chuck-dbos: Linked

docs/typescript/reference/dbos-class.md
chuck-dbos: See above - I think when we try this we will change it a bit.  `createDecorator` is not right, though.

docs/typescript/reference/dbos-class.md
chuck-dbos: So, the idea here was that these augment any existing context in certain ways, or start a new one.  We could allow use of `runWithTopContext`, this does what you want but isn't exported from `DBOS` right now.

docs/typescript/reference/dbos-class.md
chuck-dbos: PR for you: https://github.com/dbos-inc/dbos-transact-ts/pull/674

docs/typescript/reference/decorators.md
kraftp: I'd really like to get the TS docs well-structured before we officially "launch" this, right now they're hard to follow.  In my opinion that would mean a smaller number of tutorials that are totally focused on the new version. No one is going to use the new version based only on the API reference, we have to transition the tutorials and templates.

chuck-dbos: Completely agree, I think it's a second phase after this one.

docs/typescript/reference/dbos-class.md
kraftp: Agreed, let's deprioritize it. Maybe even cut this "Extensibility -> Decorators" section for now.

docs/typescript/reference/dbos-class.md
kraftp: Cut this section for now.

docs/typescript/reference/dbos-class.md
kraftp: Not a priority, honestly let's just remove the entire "Extensibility" section and come back to it later.

docs/typescript/reference/dbos-class.md
```diff
@@ -0,0 +1,971 @@
...
+```
+
+##### Using Workflow Queues
+By default, the `startWorkflow` method [described above](#starting-background-workflows) always starts the target workflow immediately.  If it is desired to control the concurrency or rate of workflow invocations, a [`queue`](../reference/workflow-queues.md) can be specified.
+
+```typescript
+DBOS.withWorkflowQueue<R>(wfq: string, callback: ()=>Promise<R>) : Promise<R>
+```
+
+The `DBOS.withWorkflowQueue` function runs the provided `callback` function and returns its result.  Any workflows started within `callback` will be run on the queue named by `wfq`.
```
kraftp, line 184: Just to be clear, this is being replaced with the new interface from https://github.com/dbos-inc/dbos-transact-ts/pull/673, right?

chuck-dbos: Augmented.

chuck-dbos: I mean, the whole neat thing about v2 is that it just looks like function calls.  That can now use queues to throttle the app pace, which is pretty cool if you think about it.

docs/typescript/reference/dbos-class.md
kraftp: Approved!

docs/typescript/reference/dbos-class.md
kraftp: I honestly don't understand the callback-based queues interface.  I can see how to use it, but I don't see why it needs a callback. The new interface based on `DBOS.startWorkflow` makes a lot more sense. Are there advantages to the callback-based interface?

chuck-dbos:
I like the plan to cut the extensibility part.  I will just commit that onto a separate branch, so it isn't lost.

docs/typescript/reference/dbos-class.md
chuck-dbos: Placed next to each other for a `startWorkflow`, no, it doesn't make much sense.
Without the startWorkflow, it's a wash, because you have the callback or you have the `await handle.getResult`
Placed apart, say as a middleware, or administrative policy, then could you see it?  Because that's where I see the value.  Run the workflows in this class on this queue, or on these HTTP routes in that queue, etc.

kraftp: I can see how you could use it, but I don't think it's enough to justify having two ways of doing it. Would rather keep things simple and support/document only the simpler/easier-to-understand way of using queues.

PR #272, opened 2024-12-16
document Cloud setup steps

maxdml:
- Document setup steps
- Add a config section to cloud application management

As we make the cloud configuration more flexible, we should make it easier for readers to read which options can be set (in the same page)

### DIFF
--- a/docs/cloud-tutorials/application-management.md
+++ b/docs/cloud-tutorials/application-management.md
@@ -16,7 +16,7 @@ To deploy your application to DBOS Cloud, run this command in its root directory
 dbos-cloud app deploy
 ```
 
-Your application is deployed using the `name` in its [dbos-config.yaml](#configuring-application-deployment).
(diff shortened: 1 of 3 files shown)


docs/cloud-tutorials/application-management.md
```diff
@@ -47,6 +49,18 @@ If you edit your application, run `dbos-cloud app deploy` again to apply the lat
 * You cannot change the database of a deployed application. You must delete and re-deploy the application.
 :::
 
+#### Configuring Application Deployment
+In `dbos-config.yaml`, you can configure your application deployment with these properties:
```
kraftp, line 53: I would add a little more detail on exactly what this does.

I would also make the recommendation to use a `build.sh` and put all the commands there so the setup is just `./build.sh`.  Doing it this way runs into incomprehensible bash escaping rules.

Finally, note that this is for pro users only.

kraftp (approved):

PR #273, opened 2024-12-16
Clarify sql_session usage

qianl15:
As pointed out by a user, we need to clarify that transaction functions shouldn't call `commit` or `rollback`. Otherwise you'll get a `sqlalchemy.exc.InvalidRequestError: Can't operate on closed transaction inside
context manager.` error.

This PR also solves some security alerts.


docs/python/reference/contexts.md
```diff
@@ -166,6 +166,11 @@ DBOS.sql_session: sqlalchemy.Session
 May only be accessed from within a transaction.
 Retrieves the SQLAlchemy session of the transaction, a database connection the transaction can use to interact with the database.
 
+:::tip
+DBOS automatically wraps your transaction function within a SQLAlchemy ["begin once" block](https://docs.sqlalchemy.org/en/20/core/connections.html#connect-and-begin-once-from-the-engine). Therefore, do not use `DBOS.sql_session.commit()` or `DBOS.sql_session.rollback()` in your functions. Otherwise, you might see a `sqlalchemy.exc.InvalidRequestError: Can't operate on closed transaction inside
+context manager` error.
```
kraftp, line 171: Does this line break in a quoted line work?

kraftp (approved):

docs/python/reference/contexts.md
qianl15: yeah it works.

PR #278, opened 2025-01-06
Refactor TS Docs

kraftp:
- Move the most important TS docs (workflows, steps, transactions...) back to the top level
- Improve and condense their text using material from the Python docs


docusaurus.config.js
```diff
@@ -154,53 +154,33 @@ const config = {
...
             to: '/typescript/tutorials/development/testing-tutorial',
           },
-          {
-            from: '/typescript/tutorials/step-tutorial',
-            to: '/typescript/tutorials/programmingmodel/step-tutorial',
-          },
           {
             from: '/typescript/tutorials/configured-instances',
-            to: '/typescript/tutorials/programmingmodel/configured-instances',
+            to: '/typescript/tutorials/development/configured-instances',
```
chuck-dbos, line 159: I don't think this belongs in the "development" section.  Maybe the point of the "development" section was missed.  The programming model bit was supposed to be about what your code says and how it works, loosely, "coding".  The "development" section was supposed to be about supporting tools and techniques.  Configured instances is a coding construct.

docs/typescript/tutorials/workflow-tutorial.md
```diff
@@ -0,0 +1,224 @@
...
+}
+```
+
+## Reliability Guarantees
+
+Workflows provide the following reliability guarantees.
+These guarantees assume that the application and database may crash and go offline at any point in time, but are always restarted and return online.
+
+1.  Workflows always run to completion.  If a DBOS process crashes while executing a workflow and is restarted, it resumes the workflow from the last completed step.
+2.  [Steps](./step-tutorial.md) are tried _at least once_ but are never re-executed after they complete.  If a failure occurs inside a step, the step may be retried, but once a step has completed, it will never be re-executed.
```
chuck-dbos, line 33: I wonder if it is worth saying something like "completed (returned a value or thrown an exception to the calling workflow)".  Some people seem to assume "completed" means "completed successfully" but it doesn't of course.

chuck-dbos (approved):

PR #286, opened 2025-01-15
Docs Update

kraftp:
Simplifying documentation using the new database wizard (https://github.com/dbos-inc/dbos-transact-py/pull/176, https://github.com/dbos-inc/dbos-transact-ts/pull/705).

Also general cleanup on the TypeScript documentation.

PR #287, opened 2025-01-15
Reference doc for launch; improve app anatomy doc

chuck-dbos:


docs/typescript/tutorials/development/application-structure-explanation.md
```diff
@@ -19,20 +19,31 @@ dbos-hello-app/
...
 │   |── operations.ts
 │   └── operations.test.ts
 ├── start_postgres_docker.js
 └── tsconfig.json
 ```
 
-The two most important files in a DBOS project are `dbos-config.yaml` and `src/operations.ts`.
-
-`dbos-config.yaml` defines the configuration of a DBOS project, including database connection information, migration configuration, and global logging configuration.
+As most of these files are common to any TypeScript / NodeJS project, the interesting file is `dbos-config.yaml`.  `dbos-config.yaml` tells DBOS how to run the program, either by running a _start_ command, or loading the program's _entrypoints_ and starting the DBOS runtime.
```
qianl15, line 29: Peter's PR updated this file https://github.com/dbos-inc/dbos-docs/pull/286
Can you rebase on that?

docs/typescript/reference/transactapi/dbos-class.md
```diff
@@ -902,6 +902,12 @@ Each field matches the pattern if its numerical value is within any of the inclu
 ## Application Lifecycle
 For DBOS applications using entrypoint classes specified in [`dbos-config.yaml`](../configuration), a "serverless" application lifecycle is handled automatically.  However, it is possible to build a custom application lifecycle.
 
+The steps are:
```
qianl15, line 905: It would be clearer to explain which function manages which part of the lifecycle (or provide some forward pointers). For example, the first three are done through `DBOS.launch`, while the last one is done through `DBOS.destroy`.
For now, it's unclear how to think about this lifecycle during app development. E.g., where should I put `DBOS.launch`?

docs/typescript/tutorials/development/application-structure-explanation.md
chuck-dbos: Significantly revised, I think a lot clearer.

docs/typescript/reference/transactapi/dbos-class.md
chuck-dbos: I've wanted a tutorial section on it, rather than a reference section.  But I will at least split it.

chuck-dbos: Made an attempt, let me know if this helps.

I think "where do I put DBOS.launch" is a question best answered by the templates.  For example, "where do I put DBOS launch in main.ts" has one answer, "where do I put DBOS.launch in a jest test" has another, and "where do I put DBOS launch in a next app" has at least two good answers, if not more.  Rather than a befuddling list of answers, get a template.  The goal here is to provide the guiding reference to what makes the templates work, or not.

docs/typescript/tutorials/development/application-structure-explanation.md
```diff
@@ -38,21 +43,31 @@ runtimeConfig:
...
+Other templates will be similar, but may have additional files under `src/`:
 ```
-It is not necessary to export classes that are already referenced by the entrypoint file(s), as these will be loaded and decorated methods will be registered.
-You can also define multiple entrypoint files using the `runtimeConfig` section of the [configuration](../../reference/configuration#runtime).
+├── src/
+│   |── main.ts
+│   └── operations.ts
+```
+
+In this structure, `src/operations.ts` is an _entrypoint_ file, where DBOS looks for event handlers, such as [scheduled workflows](../scheduled-workflows), [kafka consumers](../requestsandevents/kafka-integration), or, if you are using DBOS's built in Koa server, [request handlers](../requestsandevents/http-serving-tutorial).
```
qianl15, line 57: I thought we removed Kafka integration?

docs/typescript/reference/transactapi/dbos-class.md
```diff
@@ -909,15 +915,42 @@ DBOS.setConfig(config: DBOSConfig, runtimeConfig?: DBOSRuntimeConfig)
...
+const [cfg, rtCfg] = parseConfigFile({configfile: 'my-testing-dbos-config.yaml'});
+```
+
+Use of `parseConfigFile` allows a file other than `dbos-config.yaml` to be loaded and programmatic modifications or checks to be performed prior to calling `DBOS.setConfig`.
+
+### Loading Classes
+Before a DBOS app is launched, all classes with DBOS methods should be loaded, giving their decorators a chance to run and register the associated functions.  This is generally done automatically, but for advanced situations, it can be performed programatically with `DBOS.loadClasses`.  Note that, similar to those in the application [entrypoints](../../tutorials/development/application-structure-explanation), files provided should be the `.js` files that will actually be run during by the application.
+
+```typescript
+await DBOS.loadClasses(runtimeConfig.entrypoints);
```
qianl15, line 930: Can we have a better example than `runtimeConfig.entrypoints`? We can also make this part generic to a list of file names.. can `loadClasses` simply take a list of classes?

chuck-dbos: Better example, sure.  loadClasses always takes the files, if you already have classes then they're already loaded ;-)

kraftp (approved):

chuck-dbos:
Good catches there pk, are we ready?

PR #293, opened 2025-01-23
Documentation for Dbosignore

manojdbos:
Screenshot 2025-02-14 at 08 48 37


docs/cloud-tutorials/application-management.md
```diff
@@ -77,6 +77,43 @@ A setup script may install system packages or libraries or otherwise customize t
 apt install traceroute
 ```
 
+#### Ignoring files with .dbosignore
+
+The .dbosignore file allows you to specify files and directories that should be excluded when packaging files into a zip for deployment. It functions similarly to a .gitignore file, using patterns to define which files and directories to skip.
```
kraftp, line 82: To be honest, if our syntax is identical to `.gitignore` we can just say that and skip the rest for brevity. People know how to use `.gitignore`.

manojdbos: The request for this section was from @maxdml . @maxdml your thoughts ?

manojdbos: I'll try to condense it.

I dont think we can say it is identical to gitignore with certainity.

The best we can say is it is gitignore like and that we use glob for filtering

kraftp: Yeah, that's also fair, I just think we can condense this section. People will understand what you mean.

kraftp (approved):

PR #294, opened 2025-01-24
update queue documentations for per process concurrency

maxdml:


docs/python/reference/queues.md
```diff
@@ -13,6 +13,8 @@ Queue(
     name: str = None,
     concurrency: Optional[int] = None,
     limiter: Optional[Limiter] = None
+    *, # Disable positional arguments from here on
```
qianl15, line 16:
```suggestion
    *,
```

qianl15: I think we don't need to have this comment. It's a common thing in Python.

docs/typescript/reference/transactapi/workflow-queues.md
```diff
@@ -16,20 +16,22 @@ interface QueueRateLimit {
...
+    rateLimit?: QueueRateLimit;
+}
+
 class WorkflowQueue {
-    constructor(
-        readonly name: string,
-        readonly concurrency?: number,
-        readonly rateLimit?: QueueRateLimit
-    );
+   constructor(name: string, queueParameters: QueueParameters);
```
qianl15, line 26: Is it a breaking change or do we still support the old way?

maxdml: Not breaking -- the change is backward compatible. See https://github.com/dbos-inc/dbos-transact-ts/blob/98676abe6c8c91084243c63279ca8f706b0ed108/src/wfqueue.ts

kraftp (approved):

PR #299, opened 2025-02-01
Update Python + Typescript Programming Guides

kraftp:
Make it more direct and describe popular features like queues and scheduled workflows.

### DIFF
--- a/docs/python/programming-guide.md
+++ b/docs/python/programming-guide.md
@@ -5,8 +5,6 @@ pagination_next: python/tutorials/workflow-tutorial
 pagination_prev: quickstart
 ---
 
-import LocalPostgres from '/docs/partials/_local_postgres.mdx';
-
 This guide shows you how to use DBOS to build Python apps that are **resilient to any failure**.
 
 ## 1. Setting Up Your Environment
--- a/docs/python/programming-guide.md
+++ b/docs/python/programming-guide.md
@@ -144,10 +144,10 @@ Step two completed!
 
 You can see how DBOS **recovers your workflow from the last completed step**, executing step 1 without re-executing step 2.
 
-## 2. Queues and Parallelism
+## 3. Queues and Parallelism
 
-You can use DBOS queues to run many tasks concurrently.
-To try it out, copy this code into `main.py`:
+If you need to run many functions concurrently, use DBOS _queues_.
+To try them out, copy this code into `main.py`:
 
 ```python showLineNumbers title="main.py"
 import time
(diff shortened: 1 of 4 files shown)


docs/python/programming-guide.md
```diff
@@ -7,236 +7,386 @@ pagination_prev: quickstart
...
+Then, install DBOS and create a DBOS configuration file:
 ```shell
 pip install dbos
-dbos init
+dbos init --config
 ```
 
-Now, let's use FastAPI to write a simple app that greets our friends.
-Every time the app receives a greeting, it performs two steps:
+## 2. Workflows and Steps
```
apoliakov, line 46: What if, instead of having several code blocks, we have one code block with all the things (and a few routes like "/transaction", "/queue", "/workflow".

We show that code block at the top. Once. The advanced users see all the decorators in one place.

Then we have a few sections of text under it to describe what each thing does.... ?

kraftp: The toolbox does this!  

https://github.com/dbos-inc/dbos-demo-apps/pull/259

But for this guide that means copying and pasting a lot of code...

qianl15: We can do both. In some LangChain docs, they have a folded "see the full picture" code block. You can achieve that by using the `<details>` block.

apoliakov: ... I see the toolbox. Very cool.

... I am not sure it's that bad of an idea to have toolbox lines 12 thru 92 (80 lines) in one block of code at the top? Maybe you tried it and it was ugly? But it would only be one block of code and you could link back to the toolbox app to say "run it here"....

kraftp: Yeah, we could put the toolbox in a collapsible at the bottom

docs/python/programming-guide.md
```diff
@@ -7,236 +7,386 @@ pagination_prev: quickstart
...
-from .schema import dbos_hello
+@DBOS.workflow()
+def dbos_workflow():
+    step_one()
+    step_two()
 
-app = FastAPI()
-logging.basicConfig(level=logging.INFO)
+if __name__ == "__main__":
+    DBOS.launch()
```
qianl15, line 70: I think it'll help a lot if we describe when to call `DBOS()` (the initializer) and when to call `DBOS.launch()`. In this guide, after switching to FastAPI, the program no longer contains `DBOS.launch()`, which could be confusing. And historically, "when to init DBOS and when to use launch" is one of the most asked questions.

docs/python/programming-guide.md
```diff
@@ -7,236 +7,386 @@ pagination_prev: quickstart
...
+Step two completed!
+```
 
-# Record the greeting in the database using SQLAlchemy
-def insert_greeting(name: str) -> str:
-    with engine.begin() as sql_session:
-        query = dbos_hello.insert().values(name=name)
-        sql_session.execute(query)
-    logging.info(f">>> STEP 2: Greeting to {name} recorded in the database!")
+To see durable execution in action, let's modify the app to serve a DBOS workflow from an HTTP endpoint using FastAPI.
```
qianl15, line 87: I understand that you put FastAPI here to demonstrate recovery.
But maybe it's better to talk about FastAPI integration in a separate section. Something like "FastAPI Integration", and also explains briefly about the lifecycle hook.

docs/python/programming-guide.md
```diff
@@ -7,236 +7,386 @@ pagination_prev: quickstart
...
-Now, restart your app with `dbos start` and visit its URL again: http://localhost:8000/greeting/Mike.
-The app should durably execute your workflow the same as before!
+Every time you visit http://localhost:8000, your workflow should insert another row, and the printed row count should go up by one.
 
-The code for this guide is available [on GitHub](https://github.com/dbos-inc/dbos-demo-apps/tree/main/python/greeting-guestbook).
+Congratulations!  You've finished the DBOS Python guide.
+You can find the code from this guide in the [DBOS Toolbox](https://github.com/dbos-inc/dbos-demo-apps/tree/main/python/dbos-toolbox) template app.
 
-Next, to learn how to build more complex applications, check out our Python tutorials and [example apps](../examples/index.md).
+Next, to learn how to build more complex applications, check out the Python tutorials and [example apps](../examples/index.md).
```
qianl15, line 392: At the end, add a block that contains all code (in a details block). Then provide pointers to related tutorials pages.

docs/python/programming-guide.md
```diff
@@ -7,236 +7,386 @@ pagination_prev: quickstart
...
+Step two completed!
 ```
 
-Now, this app has a problem: if it is interrupted after signing the guestbook, but before recording the greeting in the database, then **the greeting, though sent, will never be recorded**.
-This is bad in many real-world situations, for example if a program fails to record making or receiving a payment.
-To fix this problem, we'll use DBOS durable execution.
+You can see how DBOS **recovers your workflow from the last completed step**, executing step 1 without re-executing step 2.
 
-## 2. Durable Execution with Workflows
+## 3. Queues and Parallelism
```
qianl15, line 143: Add a pointer to the queues tutorial page.

docs/python/programming-guide.md
qianl15, line 46: Also add a link to Workflows and Steps tutorials.

qianl15 (approved):

PR #304, opened 2025-02-16
Step input is pickled only when called outside a workflow

jkline-trustle:
There may be a better way to describe this.


maxdml (approved):
Thanks for the contribution!

jkline-trustle:
Another case seems to be "steps used in queues run asynchronously and somewhere else, so the input must be pickled."

kraftp:
Yes, steps called asynchronously via `start_workflow` or queues have the same issue. I'm hesitant to merge this particular change. The fact that the inputs of steps called synchronously aren't serialized is really an implementation detail, and I'm not sure if we want to expose it in our docs. I'd rather work on a more robust solution for the problems we currently work around by passing unserializable objects to steps. I'm working on some improvements to DBOS classes and configured classes that should help with this.

PR #316, opened 2025-03-06
Next.js Integration Guide

chuck-dbos:


docs/integrations/adding-dbos-to-next.md
```diff
@@ -0,0 +1,235 @@
...
+- Cron-style task scheduling – Automate recurring jobs with [cron-like scheduling](../typescript/tutorials/scheduled-workflows).
+- Background tasks and WebSockets – Keep execution and state across UI calls, with the ability to send results back to the client in real time.
+- Built-in tracing and replay debugging – [Find workflows in the dashboard](../cloud-tutorials/monitoring-dashboard) and [re-run them locally](../cloud-tutorials/timetravel-debugging).
+
+## Architectural Overview
+Next.js is a framework that optimizes where and when [React](https://react.dev/) UI components render—whether on the client, server, or edge—while also handling routing, data fetching, and performance optimizations.  Part of this architecture involves the creation of minimized code bundles for handling requests.  These bundles can be loaded quickly in a “serverless” environment, leading to minimal request latency even when the server is “cold”.  This “serverless” style of deployment precludes any long-running jobs, background tasks that execute while no client requests are pending, or long-lived server objects such as WebSockets:
+![Serverless Next.js architecture](./assets/serverlessnext.png)
```
kraftp, line 20: I like these diagrams a lot, but their header text ("Serverless Next") clips weirdly into the text above it. Probably just needs a little more padding on the top and maybe a different font.

chuck-dbos: Changed font, boxed off.  See if that helps.

docs/integrations/adding-dbos-to-next.md
```diff
@@ -0,0 +1,235 @@
+---
+sidebar_position: 40
+title: Adding DBOS to Next.js Apps
+description: Learn how to integrate DBOS into Next.js applications.
+---
+
+# Introduction
+
+## Why Add DBOS To a Next.js Application?
+[Next.js](https://nextjs.org/) is a solid choice for full-stack applications, but popular Next.js hosting options focus on serverless, CDN-heavy deployments that do not allow long-running tasks or other “heavy lifting” in the server.  By adding DBOS Transact, and running in a suitable hosting environment (such as [DBOS Cloud](https://www.dbos.dev/dbos-cloud)), the following additional features are available:
```
qianl15, line 10: I think it'll look much cleaner if we keep this list short. We can also point to the "Why DBOS" page.
- Reliable long-running tasks through lightweight durable execution. (point to workflow tutorial)
- Built-in observability and debugging support.

docs/integrations/adding-dbos-to-next.md
```diff
@@ -0,0 +1,235 @@
...
+```
+
+## Preventing Bundling
+As explained in the [Architectural Overview](#architectural-overview) above, the DBOS library, workflow functions, and related code must remain external to Next.js bundles and be loaded at startup.  If DBOS code gets bundled, the bundles will contain incomplete functions, object instances, queue definitions, etc., many of which will be duplicated across bundles.  These bundles will be loaded at runtime in response to Next.js requests, leading to a confusing mess of DBOS registration errors.
+
+Bundlers determine which code to include by tracing `import` dependencies from pages, actions, or routes and minimizing them into bundles. To prevent this, two approaches are outlined below:
+- Marking Modules as External – Configuring webpack to treat DBOS modules and files as external, ensuring they are not bundled.
+- Using globalThis – Accessing code and data through [`globalThis`](https://ja-visser.medium.com/globalreferences-in-nodejs-75f095962596) instead of import, effectively bypassing the bundler.
+
+### webpack Configuration in `next.config.ts`
```
qianl15, line 144: Maybe explicitly call the webpack config "Option 1: ..." and then the globalThis "Option 2:...". It's unclear whether a user needs both, or either one of those would work.

chuck-dbos: One, the other, both, or maybe options not listed here...

docs/integrations/adding-dbos-to-next.md
kraftp: Looks good now

docs/integrations/adding-dbos-to-next.md
```diff
@@ -0,0 +1,235 @@
...
+
+  // ... Then proceed to start the Next.js server ...
+}
+```
+
+### Adding `server.ts` To A Project
+If your project does not currently have a `server.ts` file or similar, or if you are using the default `server.js` provided by Next.js, the first step is to create one.   This file can be created in the top level of your project, inside `src/`, or in any sensible place within your app structure.
+
+The DBOS demo apps contain example `server.ts` files to copy:
+- [DBOS Next.js Template](https://github.com/dbos-inc/dbos-demo-apps/blob/main/typescript/dbos-nextjs-starter/src/server.ts): This project uses a basic `server.ts` file.
```
kraftp, line 65: I would include a simple copy-pasteable `server.ts` in a details block. The outbound links to GitHub are scary.

docs/integrations/adding-dbos-to-next.md
```diff
@@ -0,0 +1,235 @@
...
+- [DBOS Task Scheduler](https://github.com/dbos-inc/dbos-demo-apps/blob/main/typescript/nextjs-calendar/src/server.ts): This project uses a more sophisticated `server.ts` file with WebSocket support, etc.
+
+### Compilation Settings For DBOS Code
+Using DBOS with Next.js requires some TypeScript compiler settings that do not match Next.js defaults.  These can either be added to your existing `tsconfig.json`, or if they are incompatible with the existing `tsconfig.json` settings, a separate `tsconfig.server.json` file can be created for use with the server builds.
```
kraftp, line 69: This could be softened a bit, right now it sounds really scary ("uncommon settings") when it's really quite mild.

docs/integrations/adding-dbos-to-next.md
```diff
@@ -0,0 +1,235 @@
...
+  },
+```
+
+These scripts can be more sophisticated, allowing for database migrations and `nodemon`.  See the [demo apps](https://github.com/dbos-inc/dbos-demo-apps/tree/main/typescript) for examples.
+
+### DBOS Configuration
+By default, `DBOS.launch()` will read [`dbos-config.yaml`](../typescript/reference/configuration) to get key information, such as database settings.  To configure DBOS programatically, see [`DBOS.setConfig`](../typescript/reference/transactapi/dbos-class#setting-the-application-configuration).
+
+## Calling DBOS Code From Next.js
+The DBOS Transact library and application your workflow functions are available in all server-side request handling, including page loads, server actions, and route handlers.
```
kraftp, line 110: Missing some words?

docs/integrations/adding-dbos-to-next.md
kraftp: Let's try to be more prescriptive. I read this section and frankly I don't know what I need to do. Do I have to configure webpack? Do I have to do it just for the `dbos-sdk` library code or for other code as well? (can we fix Transact so it survives bundling automatically?) When do I need `globalThis`? Is it for any DBOS-decorated class?

docs/integrations/adding-dbos-to-next.md
chuck-dbos: K

docs/integrations/adding-dbos-to-next.md
chuck-dbos: Better yet, it has extra.  (Good catch.)

docs/integrations/adding-dbos-to-next.md
chuck-dbos: The "Why DBOS" page doesn't really say "why DBOS"... but I made this a lot shorter.

docs/integrations/adding-dbos-to-next.md
chuck-dbos: Transact is not broken.

chuck-dbos: Tried to make more prescriptive.

docs/integrations/adding-dbos-to-next.md
```diff
@@ -0,0 +1,287 @@
...
+`webpack` should be configured to treat the following as external dependencies:
+- The DBOS Transact library: `@dbos-inc/dbos-sdk`
+- Any DBOS [package libraries](../typescript/reference/libraries.md), such as `@dbos-inc/dbos-ses`
+- Any application source files implementing steps, transactions, or workflows
+- Any application source files that create or accesses DBOS [queues](../typescript/reference/transactapi/workflow-queues.md), [object instances](../typescript/tutorials/instantiated-objects), or other DBOS objects
+
+Ensure that each such module or file is covered in `next.config.ts`, within `webpack`'s `config.externals` array.  Entries in `config.extenals` must match the names used in `import` statements for recognition to work correctly.  It is not necessary to list each external file individually, as the `config.externals` array can also accept regular expressions and callback functions.  For example, if an application `import`s all DBOS logic with the module alias `@dbos/` (such as `import { MyWorkflow } from "@dbos/operations"`), only a single regular expression of `/^@dbos\/.+$/` is needed to treat all files as external:
```
kraftp, line 197:
```suggestion
Ensure that each such module or file is covered in `next.config.ts`, within `webpack`'s `config.externals` array.  Entries in `config.externals` must match the names used in `import` statements for recognition to work correctly.  It is not necessary to list each external file individually, as the `config.externals` array can also accept regular expressions and callback functions.  For example, if an application `import`s all DBOS logic with the module alias `@dbos/` (such as `import { MyWorkflow } from "@dbos/operations"`), only a single regular expression of `/^@dbos\/.+$/` is needed to treat all files as external:
```

kraftp (approved):
This is a good start, and it's worth merging in. However, the section on preventing bundling is still extremely complicated. We should talk about how it can be made simpler, whether through changes to Transact or a different explanation.

PR #319, opened 2025-03-14
Simplify object instances doc

chuck-dbos:
Simplifies the treatment of class instances (to be more like Python).

This removes configureInstance from the doc, as would be suitable after https://github.com/dbos-inc/dbos-transact-ts/pull/790.  There should perhaps be a waiting period ... in case someone looks at the new doc but is using an older copy of the library.

### DIFF
--- a/docs/typescript/reference/transactapi/dbos-class.md
+++ b/docs/typescript/reference/transactapi/dbos-class.md
@@ -46,29 +46,19 @@ class Workflows
 #### Decorating Instance Methods
 Workflow functions may also be instance methods.
 
-This requires a few more steps.  For instance methods to work, each class instance must be given a name and registered with DBOS, so that a corresponding instance can be found for workflow recovery.  This is done with `DBOS.configureInstance`:
-
-```typescript
-DBOS.configureInstance(cls: Constructor, name: string, ...args: unknown[]) : R
-```
-
-The arguments to `DBOS.configureInstance` are:
-* `cls`: The class to be instantiated and configured.  This class must extend `ConfiguredInstance`.
-* `name`: The name for the configured instance (which must be unique within the set of instances of the same class)
-* `...args`: Any additional arguments to pass to the constructor function `cls`.
-
-Configured classes must extend from the following abstract class:
+For instance methods to work, the class must extend from the following abstract class:
 ```typescript
 export abstract class ConfiguredInstance {
   readonly name: string;
   constructor(name: string) {
     this.name = name;
+    // ... registration occurs ...
   }
   abstract initialize(ctx: InitContext): Promise<void>;
 }
 ```
 
-The `ConfiguredInstance` base class provides naming information, and also declares the `initialize` function which will be called after the DBOS runtime has started, but before workflow processing commences.  The argument to `initialize` is an [`InitContext`](./oldapi/contexts.md#initcontext), which provides access to configuration and other information.
+The `ConfiguredInstance` base class constructor registers the instance under the given `name`.   The `initialize` function will be called on each registered instance after the DBOS runtime has started, but before workflow processing commences.  The argument to `initialize` is an [`InitContext`](./oldapi/contexts.md#initcontext), which provides access to configuration and other information.
 
 Example:
 ```typescript
@@ -76,7 +66,7 @@ import { DBOS, ConfiguredInstance } from '@dbos-inc/dbos-sdk';
 
 class MyClass extends ConfiguredInstance
 {
-  constructor(name: string) {
+  constructor(name: string, args: string[]) {
     super(name);
   }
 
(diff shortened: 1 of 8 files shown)


docs/typescript/reference/transactapi/dbos-class.md
```diff
@@ -46,37 +46,27 @@ class Workflows
 #### Decorating Instance Methods
 Workflow functions may also be instance methods.
 
-This requires a few more steps.  For instance methods to work, each class instance must be given a name and registered with DBOS, so that a corresponding instance can be found for workflow recovery.  This is done with `DBOS.configureInstance`:
-
-```typescript
```
qianl15, line 51: If we want a waiting period, we can temporarily mark this as `(Deprecated)`

chuck-dbos: I think you misunderstand what the waiting period is.

If you read the docs as proposed here, you'd be tempted to say `new Instance("name")`.  This will work with the new library.  It won't work if you grabbed your library last week ... you'll get an error about it not being registered.

qianl15: We don't version our docs (we might support it later), but yeah I mean document it as deprecated since version X.

chuck-dbos: The thing that messes up the usual deprecation process in this instance is that the new recommended syntax (`new X()`) always compiled, but now it does what you want.

We're following the usual formula, which is:
1. Add new feature.
2. Sometime later, deprecate the old, and point users toward the new.
3. Sometime later still, remove the old.

In step 2, users would generally not get very far in trying to use the new feature without noticing that they need the new version... but in this case that's possible.

docs/typescript/tutorials/instantiated-objects.md
```diff
@@ -4,96 +4,57 @@ title: Using Instantiated Objects
...
+The reason for these requirements is to enable workflow recovery.  When you create a new instance of, DBOS stores it in a global registry indexed by `name`.  When DBOS needs to recover a workflow belonging to that class, it looks up the `name` so it can run the workflow using the right class instance.  While names are used by DBOS Transact internally to find the correct object instance across system restarts, they are also potentially useful for monitoring, tracing, and debugging.
+
+## Using Configured Class Instances
+Configured class instances should be created and named when the application starts, before any workflows run.  This ensures that they will all be initialized before any processing begins.
+
+### Writing New Configured Classes
+All configured classes must:
+* Extend from the `ConfiguredInstance` base class
+* Provide a constructor, which can take any arguments, but must provide a name to the base `ConfiguredInstance` constructor
+* Have an `initialize(ctx: InitContext)` that will be called after all objects have been created, but before request handling commences
```
kraftp, line 53: Can we get rid of this requirement later?  Make it not required?

docs/typescript/tutorials/instantiated-objects.md
```diff
@@ -4,96 +4,57 @@ title: Using Instantiated Objects
...
+* Have `@DBOS.transaction`, `@DBOS.step`, and/or `@DBOS.workflow` methods to be called on the instances
```
kraftp, line 54: We can delete this line, it's not really required so much as the whole reason you're doing this

docs/typescript/tutorials/instantiated-objects.md
```diff
@@ -4,96 +4,57 @@ title: Using Instantiated Objects
...
+
 ### `initialize()` Method
 The `initialize(ctx: InitContext)` method will be called during application initialization, after the code modules have been loaded, but before request and workflow processing commences.  The `InitContext` argument provides configuration file, logging, and database access services, so any validation of connection information (complete with diagnostic logging and reporting of any problems) should be performed in `initialize()`.
 
 ## Notes
-As `@DBOS.getApi`, `@DBOS.putApi`, and similar handler registration decorators specify the URL directly, it does not make sense to use these on configured class instances, as there is no way to tell which instance is to handle the request.
-
-The name of a workflow, and the name of the configuration in use, is kept in the DBOS system database so that interrupted workflows can be resumed.  It is therefore important to keep names consistent across application deployments, unless there is no chance that a pending workflow will need to be recovered.
+As `@DBOS.scheduled`, `@KafkaConsume`, `@DBOS.getApi`, `@DBOS.putApi`, and similar handler registration decorators cannot be applied to instance methods.
```
kraftp, line 60:
```suggestion
Handler registration decorators such as `@DBOS.scheduled`, `@KafkaConsume`, `@DBOS.getApi`, and `@DBOS.putApi` cannot be applied to instance methods.
```

docs/typescript/tutorials/instantiated-objects.md
```diff
@@ -4,96 +4,57 @@ title: Using Instantiated Objects
 description: Learn how to make workflows, transactions, and steps reusable and configurable by instantiating objects
```
kraftp, line 4: Let's rename this guide something like "Working with TypeScript Classes", echoing Python?  No one will know what "instantiated objects" are.

docs/typescript/reference/transactapi/dbos-class.md
kraftp: I'm happy to just delete it entirely, keep the docs simple.

kraftp (approved):
Much clearer now!

PR #324, opened 2025-03-20
Test doc adjustments

chuck-dbos:
Bring testing docs in line w/ existing example code.
A bigger adjustment can be made later...


kraftp (commented):
Looks good, but don't we want a note about the new `dropSystemDB` capability?

chuck-dbos:
> Looks good, but don't we want a note about the new `dropSystemDB` capability?

Yes.  But it's in a bit of limbo given the questions about `setConfig` (you can't drop the system DB after launch, but it needs the config Max wants to pass in at launch), and it's not in the example we're claiming to be describing... or in any example yet, just our internal tests...

kraftp (approved):

PR #329, opened 2025-03-26
Small conductor touchups

chuck-dbos:


kraftp (approved):
Yeah these are really good improvements, made some small tweaks

PR #334, opened 2025-03-31
Decorator and Context split

chuck-dbos:
Split the part of decorators and contexts that are "v1" and are fully replaced from those that are not.

Suggestions for those that are not:
1. Leave alone - always an option.
2. For InitContext, this can be optional in all places where it is used, and `DBOS.` accesses can be used instead.
3. With the exception of `SkipLogging`, we have a better strategy for the parameter decorators... but it's not been implemented yet.


kraftp (commented):
I don't understand this PR. We create a new section called "Decorators" but it doesn't contain any documentation for our actual core decorators like `@DBOS.workflow`. I think we should split the current DBOS Class page in three, just like in Python:
- DBOS Class (focusing on lifecycle)
- Decorators (but all the decorators, not just some of them)
- DBOS Methods & Variables (all the `DBOS.` methods and variables)

chuck-dbos:
> I don't understand this PR. We create a new section called "Decorators" but it doesn't contain any documentation for our actual core decorators like `@DBOS.workflow`. I think we should split the current DBOS Class page in three, just like in Python:
> 
> * DBOS Class (focusing on lifecycle)
> * Decorators (but all the decorators, not just some of them)
> * DBOS Methods & Variables (all the `DBOS.` methods and variables)

That's a fair goal but the point of the exercise was to split the existing decorators page into deprecated vs not.  How to structure the DBOS class doc is a different task entirely.

chuck-dbos:
We create a new section called "Decorators" - actually we just moved it out of the page called "oldapi"

kraftp (approved):

PR #338, opened 2025-04-05
DBOS Prompts

kraftp:


docs/typescript/prompting.md
```diff
@@ -0,0 +1,814 @@
...
+
+#### setEvent
+
+Any workflow can call `DBOS.setEvent` to publish a key-value pair, or update its value if has already been published.
+ONLY call this from a workflow function, NEVER from a step.
+
+```typescript
+DBOS.setEvent<T>(key: string, value: T): Promise<void>
+```
+#### get_event
```
chuck-dbos, line 328: getEvent

docs/typescript/prompting.md
```diff
@@ -0,0 +1,814 @@
...
+
+```typescript
+DBOS.sleep(durationMS: number): Promise<void>
+```
+
+* DBOS.sleep: sleep for `durationMS` milliseconds.
+
+These functions work in any context, and will use the system sleep if no workflow is in progress.
+
+### Configurable Retries
```
chuck-dbos, line 438: Deprecate.  Run a retry loop for things that need it, paying attention to retriable and non-retriable errors.  The AI is a lot more likely to know what this means in the user app than anything we can possibly write.

qianl15 (approved):
PR is a great start! We'll keep tweaking it. For now, we can add an instruction to teach how to block forever if the program only contains scheduled workflows.

PR #340, opened 2025-04-07
Cloud cli updates

maxdml:
-  Do not advertise using DBOS Cloud DB for the "local postgres" guide
- Remove reference to rollback command
- Update db connect CLI documentation


qianl15 (approved):

docs/partials/_local_postgres.mdx
```diff
@@ -2,25 +2,6 @@ import CodeBlock from '@theme/CodeBlock';
 
 <Tabs groupId="postgres-or-docker" className="small-tabs">
```
kraftp, line 3: Is this box still used anywhere? If it is, we should remove it from there, then delete it entirely.

maxdml: It wasn't -- it was still referenced (but unused) in some demos apps. I removed the reference and remove it entirely, alongside "QuickStartDeploy" which was equally unused.

maxdml: see https://github.com/dbos-inc/dbos-docs/pull/340/commits/e3d033b57eacd17d08d47a1aa567bd1caabf8c63

kraftp (approved):

PR #351, opened 2025-04-16
More Config Updates

kraftp:
Remove all mentions of the DB Wizard, explicit postgres start commands.

### DIFF
--- a/docs/python/integrating-dbos.md
+++ b/docs/python/integrating-dbos.md
@@ -8,11 +8,17 @@ This guide shows you how to add the open-source [DBOS Transact](https://github.c
 
 ### 1. Install DBOS
 `pip install` DBOS into your application.
-Then, create a DBOS configuration file (this file is optional but is used by tooling like the DBOS CLI and debugger).
 
 ```shell
 pip install dbos
-dbos init my-app --config
+```
+
+DBOS requires a Postgres database.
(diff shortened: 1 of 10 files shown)


docs/python/integrating-dbos.md
```diff
@@ -8,11 +8,17 @@ This guide shows you how to add the open-source [DBOS Transact](https://github.c
...
 `pip install` DBOS into your application.
-Then, create a DBOS configuration file (this file is optional but is used by tooling like the DBOS CLI and debugger).
 
 ```shell
 pip install dbos
-dbos init my-app --config
+```
+
+DBOS requires a Postgres database.
+If you already have Postgres, you can set the `DBOS_DATABASE_URL` environment variable to your connection string.
```
qianl15, line 17: I don't think setting `DBOS_DATABASE_URL` alone will connect to the correct database. You still need to pass it in to the config. It works now because our template app contains the code to read `DBOS_DATABASE_URL` and pass it into the config object.

Are we changing the behavior? If so, I don't see it in the code.

qianl15, line 17:
```suggestion
If you already have Postgres, you can set the `DBOS_DATABASE_URL` environment variable to your connection string (in a later step, we'll pass it into DBOS).
```

qianl15 (approved):

PR #355, opened 2025-04-18
Document Timeouts

kraftp:


docs/python/reference/contexts.md
```diff
@@ -540,6 +540,33 @@ with SetWorkflowID("very-unique-id"):
...
+    workflow_timeout_sec: Optional[float]
+)
+```
+
+Set a timeout for all enclosed workflow invocations or enqueues.
+When the timeout expires, the workflow **and all its children** are cancelled.
+Cancelling a workflow sets its status to `CANCELLED` and preempts its execution at the beginning of its next step.
+
+Timeouts are **start-to-completion**: if a workflow is enqueued, the timeout does not begin until the workflow is dequeued and starts execution.
+Also, timeouts are **durable**: they are stored in the database and persist across restarts, so workflows can have very large timeouts.
```
apoliakov, line 556:
```suggestion
Also, timeouts are **durable**: they are stored in the database and persist across restarts, so workflows can have very long timeouts.
```

apoliakov: (two more below it looks like)

apoliakov (approved):

docs/python/reference/contexts.md
```diff
@@ -540,6 +540,33 @@ with SetWorkflowID("very-unique-id"):
     example_workflow()
 ```
 
+### SetWorkflowTimeout
```
maxdml, line 543: Let's also document detaching a workflow from its parent timeout

qianl15 (approved):

PR #358, opened 2025-04-23
[docs] Fix TS examples using str instead of string

govindpuff:
### Description of the fix/feature
Some Typescript code examples used `str` instead of `string` for string arguments. Also fixes this in the Typescript AI prompt.

### Brief description of implementation
n/a


qianl15 (approved):

qianl15:
Nice catch! Thanks for fixing them.

PR #368, opened 2025-05-05
Python v1 config

maxdml:
- Document `DBOSConfig.database_url`
- Document `DBOSConfig.otlp_logs_endpoints`
- Document `dbos-cloud app change-database-instance`
- Document `--db-url` flag for `dbos migrate`

### DIFF
--- a/docs/production/dbos-cloud/application-management.md
+++ b/docs/production/dbos-cloud/application-management.md
@@ -27,8 +27,7 @@ If your account is part of an [organization](./account-management.md#organizatio
 :::tip
 * Applications should serve requests from port 8000 (Python&mdash;the default port for FastAPI and Gunicorn) or 3000 (TypeScript&mdash;the default port for Express and Koa).
 * Multiple applications can connect to the same Postgres database server&mdash;they are deployed to isolated databases on that server.
-* You cannot change the _logical_ database of a deployed application. You must delete and re-deploy the application.
-* But you can change the database _instance_ of a deployed application with `dbos-cloud app change-database-instance`.
+* To change the database server of a deployed application, use `dbos-cloud app change-database-instance`.
 :::
 
 #### Applications Configuration
--- a/docs/production/dbos-cloud/application-management.md
+++ b/docs/production/dbos-cloud/application-management.md
@@ -27,7 +27,8 @@ If your account is part of an [organization](./account-management.md#organizatio
 :::tip
 * Applications should serve requests from port 8000 (Python&mdash;the default port for FastAPI and Gunicorn) or 3000 (TypeScript&mdash;the default port for Express and Koa).
 * Multiple applications can connect to the same Postgres database server&mdash;they are deployed to isolated databases on that server.
-* You cannot change the database of a deployed application. You must delete and re-deploy the application.
+* You cannot change the _logical_ database of a deployed application. You must delete and re-deploy the application.
+* But you can change the database _instance_ of a deployed application with `dbos-cloud app change-database-instance`.
 :::
 
 #### Applications Configuration
--- a/docs/python/reference/configuration.md
+++ b/docs/python/reference/configuration.md
@@ -24,28 +24,33 @@ class DBOSConfig(TypedDict):
     name: str
     database_url: Optional[str]
     sys_db_name: Optional[str]
-    app_db_pool_size: Optional[int]
     sys_db_pool_size: Optional[int]
     db_engine_kwargs: Optional[Dict[str, Any]]
     log_level: Optional[str]
     otlp_traces_endpoints: Optional[List[str]]
-    run_admin_server: Optional[bool]
+    otlp_logs_endpoints: Optional[List[str]]
     admin_port: Optional[int]
+    run_admin_server: Optional[bool]
 ```
 
 - **name**: Your application's name.
-- **database_url**: A connection string to a Postgres database. The supported format is
+- **database_url**: A connection string to a Postgres database. DBOS will use the connection string, unmodified, to create a [SQLAlchemy engine](https://docs.sqlalchemy.org/en/20/core/engines.html).
+
+:::info
+Passwords in connection strings should be escaped (e.g., with [urllib](https://docs.python.org/3/library/urllib.parse.html#urllib.parse.quote)).
+:::
+
+:::info
+DBOS Transact requires the [psycopg 3](https://www.psycopg.org/psycopg3/docs/) driver and will accordingly set the driver to `postgresql+psycopg` before creating an engine.
+:::
+
+I none is provided, DBOS will create a default connection string:
 ```
-postgresql://[username]:[password]@[hostname]:[port]/[database name]
+postgresql://postgres:${PGPASSWORD}@localhost:5432/application_name?connect_timeout=10&sslmode=prefer
 ```
-The `sslmode=require`, `sslrootcert` and `connect_timeout` parameter keywords are also supported.
-Defaults to:
-```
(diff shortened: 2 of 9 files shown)


docs/production/dbos-cloud/application-management.md
```diff
@@ -27,7 +27,8 @@ If your account is part of an [organization](./account-management.md#organizatio
 :::tip
 * Applications should serve requests from port 8000 (Python&mdash;the default port for FastAPI and Gunicorn) or 3000 (TypeScript&mdash;the default port for Express and Koa).
 * Multiple applications can connect to the same Postgres database server&mdash;they are deployed to isolated databases on that server.
-* You cannot change the database of a deployed application. You must delete and re-deploy the application.
+* You cannot change the _logical_ database of a deployed application. You must delete and re-deploy the application.
```
kraftp, line 30: Regular reminder: "logical database" isn't a real thing, it's a term we made up

docs/python/reference/configuration.md
```diff
@@ -24,29 +24,34 @@ class DBOSConfig(TypedDict):
...
+
+:::info
+Passwords in connection strings should be escaped (e.g., with [urllib](https://docs.python.org/3/library/urllib.parse.html#urllib.parse.quote)).
+:::
+
+:::info
+DBOS Transact requires the [psycopg 3](https://www.psycopg.org/psycopg3/docs/) driver and will accordingly set the driver to `postgresql+psycopg` before creating an engine. If you use [alembic](https://alembic.sqlalchemy.org/en/latest/) to manage database migrations, you'll need to set the driver in your connection string.
+:::
+
+I none is provided, DBOS will use this default connection string:
```
kraftp, line 47:
```suggestion
If none is provided, DBOS will use this default connection string:
```

docs/python/reference/configuration.md
```diff
@@ -24,29 +24,34 @@ class DBOSConfig(TypedDict):
...
 ```
-postgresql://postgres:dbos@localhost:5432/[application name]
+postgresql://postgres:${PGPASSWORD}@localhost:5432/application_name?connect_timeout=10&sslmode=prefer
 ```
-- **sys_db_name**: Name for the [system database](../../explanations/system-tables) in which DBOS stores internal state. Defaults to `{database name}_dbos_sys`.
-- **app_db_pool_size**: The size of the connection pool used by [transactions](../tutorials/transaction-tutorial.md) to connect to your application database. Defaults to 20.
-- **sys_db_pool_size**: The size of the connection pool used for the [DBOS system database](../../explanations/system-tables). Defaults to 20.
-- **db_engine_kwargs**: Additional keyword arguments passed to SQLAlchemy’s [`create_engine()`](https://docs.sqlalchemy.org/en/20/core/engines.html#sqlalchemy.create_engine), applied to both the application and system database engines. Defaults to:
+
+If `PGPASSWORD` is not exported, the default password will be `dbos`. The database name will be derived from your application's name.
```
kraftp, line 52: Can you change this back to what it was before?  Much clearer I think. Also you don't need to specify sslmode=prefer, that's the default.

docs/python/reference/configuration.md
```diff
@@ -55,9 +60,20 @@ postgresql://postgres:dbos@localhost:5432/[application name]
   "connect_args": {"connect_timeout": 10}
 }
 ```
-If `app_db_pool_size` or `sys_db_pool_size` is explicitly set, it will override the `pool_size` value for the respective database.
+
+:::info
+In SQLAlchemy, `connect_timout` set in `engine_kwargs` takes precedence over `connect_timeout` set in a connection string.
+
+If you set `connect_timeout` in your connection string but not in `engine_kwargs`, DBOS will set this `connect_timeout` for you in `engine_kwargs`.
```
kraftp, line 67: Why do we do this? Why does this need to be a documented part of our interface?

docs/python/reference/configuration.md
```diff
@@ -24,29 +24,34 @@ class DBOSConfig(TypedDict):
...
-The `sslmode=require`, `sslrootcert` and `connect_timeout` parameter keywords are also supported.
-Defaults to:
+- **database_url**: A connection string to a Postgres database. DBOS will use the connection string, unmodified, to create a [SQLAlchemy engine](https://docs.sqlalchemy.org/en/20/core/engines.html).
+
+:::info
+Passwords in connection strings should be escaped (e.g., with [urllib](https://docs.python.org/3/library/urllib.parse.html#urllib.parse.quote)).
+:::
+
+:::info
+DBOS Transact requires the [psycopg 3](https://www.psycopg.org/psycopg3/docs/) driver and will accordingly set the driver to `postgresql+psycopg` before creating an engine. If you use [alembic](https://alembic.sqlalchemy.org/en/latest/) to manage database migrations, you'll need to set the driver in your connection string.
```
kraftp, line 44: Why does the second sentence need to be there?  Of COURSE if you use Alembic you need to set a driver, what we we changing?

docs/python/tutorials/logging-and-tracing.md
```diff
@@ -33,6 +33,10 @@ For example, if a FastAPI HTTP endpoint calls a workflow that calls a transactio
 The transaction span is a child of the workflow span, which is a child of the HTTP endpoint span.
 You can access your current span via [`DBOS.span`](../reference/contexts.md#span).
 
+### Logging
```
kraftp, line 36: Why is this not in the existing logging section above?

docs/python/reference/configuration.md
```diff
@@ -24,29 +24,34 @@ class DBOSConfig(TypedDict):
...
-    run_admin_server: Optional[bool]
+    otlp_logs_endpoints: Optional[List[str]]
     admin_port: Optional[int]
+    run_admin_server: Optional[bool]
 ```
 
 - **name**: Your application's name.
-- **database_url**: A connection string to a Postgres database. The supported format is
-```
-postgresql://[username]:[password]@[hostname]:[port]/[database name]
```
kraftp, line 39: Why did we remove this?

docs/python/reference/configuration.md
```diff
@@ -24,29 +24,34 @@ class DBOSConfig(TypedDict):
...
-```
-The `sslmode=require`, `sslrootcert` and `connect_timeout` parameter keywords are also supported.
-Defaults to:
+- **database_url**: A connection string to a Postgres database. DBOS will use the connection string, unmodified, to create a [SQLAlchemy engine](https://docs.sqlalchemy.org/en/20/core/engines.html).
+
+:::info
+Passwords in connection strings should be escaped (e.g., with [urllib](https://docs.python.org/3/library/urllib.parse.html#urllib.parse.quote)).
```
kraftp, line 40: SQLAlchemy requires...

docs/python/reference/configuration.md
maxdml: Before we were supporting an incomplete format (w/o query parameters), that's why we needed to break down the format we support. Now we support arbitrary connection strings.

docs/python/reference/configuration.md
maxdml: I found a somewhat un-intuitive in our current behavior: if you set your `DBOS_DATABASE_URL` w/o the driver, it'll fail for the migration but work otherwise.

I thought about making the driver mandatory in the user-provided connection string. But such a connection string won't work with other tools like `psql`.

docs/python/reference/configuration.md
maxdml: Because we now support arbitrary connection strings, not the subset above, which had to be detailed. If you think this will help users, I can add it back

kraftp (approved):

PR #384, opened 2025-06-09
fix: list item markdown in custom-event-receiver.md

joebowbeer:
Add space after asterisks for list item formatting.


qianl15:
Thanks for your PR! We're re-writing this section soon so let's close this PR for now.

PR #390, opened 2025-07-01
Python Docs Improvement

kraftp:


docs/typescript/integrating-dbos.md
```diff
@@ -7,10 +7,14 @@ This guide shows you how to add the open-source [DBOS Transact](https://github.c
...
 :::info
 Also check out the integration guides for popular TypeScript frameworks:
-- [Next.js + DBOS](../integrations/adding-dbos-to-next.md)
 - [Nest.js + DBOS](../integrations/nestjs.md)
 :::
 
+:::warning
+DBOS currently is not compatible with projects that use JavaScript or TypeScript bundlers.
+Support will be added in an upcoming release.
+:::
```
chuck-dbos, line 16: This isn't clear / right.  Better to say that the DBOS library and DBOS workflows cannot be bundled with the bundler (which is not to say that your project can't use a bundler in the project at all).  We're probably not able to "add support" for some kinds of bundling, either but we can possibly add support for bundling the app as a single file in the future... possibly.

kraftp: Sure, rephrased

qianl15 (approved):

PR #402, opened 2025-08-05
Document DS and event receiver v3 APIs

chuck-dbos:
Add a (quite technical) page on extending DBOS via
- Datasources
- External event receivers

### DIFF
--- a/docs/python/reference/contexts.md
+++ b/docs/python/reference/contexts.md
@@ -42,8 +42,9 @@ DBOS.start_workflow_async(
 ) -> Coroutine[Any, Any, WorkflowHandleAsync[R]]
 ```
 
-Start an asynchronous workflow in the background and return a [handle](./workflow_handles.md) to it.
+Start an asynchronous workflow and return a [handle](./workflow_handles.md) to it.
 The `DBOS.start_workflow_async` method resolves after the handle is durably created; at this point the workflow is guaranteed to run to completion even if the app is interrupted.
+The workflow started with `DBOS.start_workflow_async` runs in the same event loop as its caller.
 
 **Example syntax:**
 
@@ -52,7 +53,7 @@ The `DBOS.start_workflow_async` method resolves after the handle is durably crea
 async def example_workflow(var1: str, var2: str):
     DBOS.logger.info("I am a workflow")
 
-# Start example_workflow in the background
+# Start example_workflow
 handle: WorkflowHandleAsync = await DBOS.start_workflow_async(example_workflow, "var1", "var2")
 ```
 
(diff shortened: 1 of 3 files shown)


docs/typescript/reference/plugins.md
```diff
@@ -0,0 +1,172 @@
...
+DBOS Transact for TypeScript currently provides two explicit extension mechanisms:
+* [Datasources](#datasources) integrate database clients, ORMs, or other resouces with the DBOS lifecycle and transaction functions.
+* [External Event Receivers](#event-receivers) integrate event receivers (pollers, inbound sockets, timers, etc.) with the DBOS lifecycle, and initate workflow functions upon receiving events.
+
+## Datasources
+
+### Method Of Operation
+
+The principle behind DBOS datasources is that DBOS and the datasource work together to execut functions in a checkpointed, transactional context.
+1. App code calls a transaction function (which may or may not be registered in advance).  Generally, this call is made to the datasource instance, as it provides full type checking for the transaction configuration.
```
devhawk, line 21:
```suggestion
1. App code calls a transaction function (which does not have to be registered in advance).  Generally, this call is made to the datasource instance, as it provides full type checking for the transaction configuration.
```

docs/typescript/reference/plugins.md
```diff
@@ -0,0 +1,172 @@
...
+ - `DBOSDataSource`: User-facing datasource API.  Part interface and part guideline, this is technically just a suggestion, but following it ensures a consistent developer experience across datasources.
+ - `DataSourceTransactionHandler`: This interface is provided by the datasource to DBOS, so that DBOS can control transaction scope.
+ - `@dbos-inc/dbos-sdk/datasource`: Exports from this submodule are used by datasources to do register themselves and initiate transactions.
+
+#### `DBOSDataSource`
+
+`DBOSDataSource` is the API for datasources to present to the application developer.  It provides a consistent, type-safe way to register and call transaction functions.
+`DBOSDataSource` is templatized on its `Config` type, which describes datasource-specific configuration for a transaction function.  `Config` typically covers transaction isolation, read-write/read-only, and the transaction's name.
+
+ - `runTransaction` should run its callback, generally with the `runTransaction` function [below](#dbos-incdbos-sdkdatasource)
```
devhawk, line 41: I don't like how we keep saying "should" and "generally". Maybe something more like: The `DBOSDataSource.runTransaction` method is used to execute arbitrary database code and save the return value in a single transaction. DBOS provides the `runTransaction` helper function described below that can be used when implementing `DBOSDataSource.runTransaction`.

docs/typescript/reference/plugins.md
```diff
@@ -0,0 +1,172 @@
...
+    external: AnyConstructor | object | string,
+    func: (this: This, ...args: Args) => Promise<Return>,
+    target: FunctionName & {
+      param: number | string;
+    },
+  ) : object | undefined;
+```
+
+The parameters to these functions are:
+ - `external`: A key that identifies the event reciever; this can be its constructor, an instance, or a name string.
```
devhawk, line 128:
```suggestion
 - `external`: A key that identifies the event receiver; this can be its constructor, an instance, or a name string.
```

PR #470, opened 2025-11-20
Queue Levels

kraftp:


docs/python/tutorials/queue-tutorial.md
```diff
@@ -200,6 +200,41 @@ def on_user_task_submission(user_id: str, task: Task):
...
+Enqueue a "concurrency manager" workflow to the partitioned queue, which then enqueues your actual workflow
+to the non-partitioned queue and awaits its result.
+This ensures both queues' flow control limits are enforced on your workflow.
+For example:
+
+```python
+# By using two levels of queueing, we enforce both a concurrency limit on each partition
+# and a global worker concurrency limit of 5, meaning that no more than 5 tasks across all
+# partitions can run on any single server.
+concurrency_queue = Queue("concurrency-queue", worker_concurrency=5)
```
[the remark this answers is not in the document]
kraftp: I'll change the example to make it clearer, but this is correct as written (and if you want worker concurrency on a partitioned queue, this is how)

docs/python/tutorials/queue-tutorial.md
```diff
@@ -200,6 +200,41 @@ def on_user_task_submission(user_id: str, task: Task):
...
+    # The "concurrency manager" workflow enqueues the process_task
+    # workflow on the non-partitioned queue and awaits its results
+    # to enforce global flow control limits.
+    return concurrency_queue.enqueue(process_task, task).get_result()
+
+@DBOS.workflow()
+def process_task(task):
+    ...
+```
+
```
maxdml: The handler should first enqueue `concurrency_manager` in the `concurrency_queue` and that workflow should dispatch the task in a `partitioned_queue` partition. This code is dispatching the task straight to `partitioned_queue`

kraftp: I see, I copy-pasted it wrong haha--I do have a working example

docs/python/tutorials/queue-tutorial.md
```diff
@@ -200,6 +200,41 @@ def on_user_task_submission(user_id: str, task: Task):
...
+Enqueue a "concurrency manager" workflow to the partitioned queue, which then enqueues your actual workflow
+to the non-partitioned queue and awaits its result.
+This ensures both queues' flow control limits are enforced on your workflow.
+For example:
+
+```python
+# By using two levels of queueing, we enforce both a concurrency limit of 1 on each partition
+# and a global concurrency limit of 5, meaning that no more than 5 tasks can run concurrently
+# across all partitions (and at most one task per partition).
+concurrency_queue = Queue("concurrency-queue", concurrency=5)
```
qianl15: This works, but also to balance load, we can set worker_concurrency on this queue too.

PR #474, opened 2025-12-03
document cybersecurity agent

maxdml:


kraftp (commented):
I have no problem with using AI to generate code as long as it's carefully reviewed, but writing is different. If we're asking our users to read something, we should put real effort into writing it as clearly and concisely as possible.

Things like the "benefits" or "key features" bullet points are not how real developers write, and don't convey useful information, and readers will know that.

maxdml:
> I have no problem with using AI to generate code as long as it's carefully reviewed, but writing is different. If we're asking our users to read something, we should put real effort into writing it as clearly and concisely as possible.
> 
> Things like the "benefits" or "key features" bullet points are not how real developers write, and don't convey useful information, and readers will know that.

These bullet points I find convenient to summarize what's going on in the code. Some programmers (including me) would rather have them first rather than diving straight into an ocean of `if err := nil`.

I think what we could do is highlight the relevant piece of codes that match each bullet point, so at least there is a visual cue as to what the reader should focus on. Another option is to drop some of the actual code, but that can be even more confusing.

PR #478, opened 2025-12-05
Queue Patterns

kraftp:


docs/python/examples/queue-patterns.md
```diff
@@ -0,0 +1,117 @@
+---
+displayed_sidebar: examplesSidebar
+sidebar_position: 36
+title: Advanced Queue Patterns
+---
+
+This example demonstrates how to use several advanced queue patterns in DBOS.
```
qianl15: Maybe also add one example for priority? You can queue a lot of tasks but assign priority to each one.

PR #513, opened 2026-02-05
Simplify docs for OpenAI Agents SDK integration

qianl15:


maxdml (approved):

qianl15:
After discussion, decided to drop this PR because it'll make upgrading DBOS easier if we install DBOS explicitly.

PR #543, opened 2026-04-20
Update hosting-conductor page to reflect upcoming conductor identities

lanice:
Actually only very minor changes needed, the env variables are mostly the same.


lanice:
@maxdml You will update the kubernetes page once you are done with that if need be I assume?

kraftp (approved):

PR #548, opened 2026-04-28
Java 0.8 updates

devhawk:


docs/java/tutorials/scheduled-workflows.md
```diff
@@ -4,22 +4,124 @@ title: Scheduled Workflows
...
+
+## Triggering a Schedule Immediately
+
+To fire a scheduled workflow immediately outside its normal cadence:
+
+```java
+WorkflowHandle<?, ?> handle = dbos.triggerSchedule("daily-report");
+```
+
+## Using the `@Scheduled` Annotation
```
kraftp: I forgot this was still there--while we're making breaking changes, I'd just remove it entirely from both docs and code.

devhawk: Would have been good to know this when working on https://github.com/dbos-inc/dbos-transact-java/pull/331...

But yeah no problem

docs/java/tutorials/spring-boot-starter.md
```diff
@@ -0,0 +1,144 @@
+---
+sidebar_position: 90
+title: Spring Boot Integration
+description: Add DBOS durable workflows to a Spring Boot application.
+---
+
+The `transact-spring-boot-starter` integrates DBOS into a Spring Boot application with zero boilerplate: add the dependency, configure a datasource, annotate your beans, and DBOS launches automatically alongside the Spring context.
```
kraftp: It's not urgent, but this should be a general integration guide instead of a guide built around a particular starter app. Wouldn't be a huge change as most of the page is already a general integration guide.

docs/java/tutorials/scheduled-workflows.md
kraftp: Yeah, my bad, I just completely forgot we had the old interface in Java

kraftp (approved):
This looks great! I can't be sure we didn't miss anything, but the new pages look good and the new API improves the examples.

docs/java/tutorials/spring-boot-starter.md
devhawk: other than the sample code, what about this is specific to the widget store demo? Do you just want the sample code to be `class Example` similar to other docs

kraftp: Basically, and mention the starter in passing instead of leading with it ("Also check out the fully worked example on GitHub...")

PR #551, opened 2026-05-06
Updated DBOS.listen_queues signature to reflect changed type annotation

pmalic:
Follow up to dbos-inc/dbos-transact-py#662


kraftp (approved):
Thanks!

PR #563, opened 2026-05-27
Fix self bean name

varunu28:
Self reference should be of the same type as that of the service i.e. `OrderService` & not of type `WidgetStoreService`. This PR fixes the documentation to specify correct bean name for `self`


devhawk (approved):
good catch. Thank you!

PR #568, opened 2026-05-29
Python + TS Update

kraftp:


docs/python/reference/contexts.md
```diff
@@ -123,11 +125,60 @@ DBOS.send_async(
...
+    topic: Optional[str] = None
+    # If set, the message is sent only once no matter how many times it is submitted with this key.
+    idempotency_key: Optional[str] = None
+```
+
+The send is atomic: if any message cannot be delivered (for example, its destination workflow does not exist), the entire batch is rolled back and no messages are sent.
+The `send_bulk` function should not be used in [coroutine workflows](../tutorials/workflow-tutorial.md#coroutine-async-workflows), [`send_bulk_async`](#send_bulk_async) should be used instead.
+
+**Parameters:**
+- `messages`: The list of `SendMessage` objects to send. Two messages in the same call may not share an idempotency key.
```
qianl15, line 165: Do we want to clarify that the idempotency key is scoped per destination ID?

qianl15 (approved):

PR #570, opened 2026-06-06
Improve Conductor Docs

kraftp:


docs/golang/programming-guide.md
```diff
@@ -347,6 +347,36 @@ Successfully completed 10 tasks
 You can see how all ten steps run concurrently&mdash;even though each takes five seconds, they all finish at the same time.
 Learn more about DBOS queues [here](./tutorials/queue-tutorial.md).
 
+## 4. Connecting to Conductor
```
qianl15: I think we might want to always call it "DBOS Conductor" because Conductor itself is a very overloaded word, including other workflow engines.

qianl15 (approved):

PR #579, opened 2026-06-23
DBOSify

kraftp:


docs/explanations/dbosify.md
```diff
@@ -0,0 +1,115 @@
...
+client = await Client.connect(
+    system_database_url,    # Postgres connection string
+    namespace="default",    # optional; each namespace maps to its own Postgres schema
+)
+```
+
+The returned client owns its underlying connection. Close it with `await client.close()` or by using it as an `async with` context manager.
+(For full control, build a `dbos.DBOSClient` yourself and pass it to the `Client(...)` constructor.)
+
+**Worker.** To configure a `Worker` pass it a Postgres connection string or a `dbos.DBOSConfig`:
```
qianl15: I assume closing a worker will also close the Postgres connection?

docs/explanations/dbosify.md
qianl15, line 1: Not sure if we want to put this under "explanations". Maybe we can start a new section for all the migration guides.

qianl15 (approved):

docs/explanations/dbosify.md
kraftp: This is a larger change, but yeah, we could move this migration/comparison material later

PR #605, opened 2026-08-13
Document the Conductor API and the dbosctl CLI

devhawk:
Conductor's HTTP API and the `dbosctl` CLI had no documentation. This adds two pages under **Deploy To Production** and links them from the Conductor overview.

Both are new files, so the diff is additive — nothing existing is rewritten beyond one bullet on `conductor.md`.

## The pages

**[Conductor API](docs/production/conductor-api.md)** — base URLs, the three ways to get the OpenAPI spec, authentication, resource scoping, the RFC 9457 error model, listing and filtering conventions, a map of all 64 operations by area, the operations a no-auth self-hosted deployment does not register, and how to generate a client.

**[dbosctl CLI Reference](docs/production/dbosctl.md)** — installation, profiles, authentication, configuration precedence, output formats, exit codes, and a command-by-command reference. Follows the shape of the existing Cloud CLI reference page.

The sidebar is autogenerated from the directory, so no `sidebars.js` change is needed.

## Where the content came from

Not from the READMEs. The endpoint reference, parameter lists, and error shapes were read out of the OpenAPI spec, and the CLI reference was generated from a recursive `--help` dump of a real build, then diffed back against it mechanically — 38 leaf commands, no undocumented commands, no invented flags.

Details that came from reading the implementation rather than the spec text, because a reader would otherwise hit them as surprises:

- **Workflow mutations need a live executor.** Cancel, resume, restart, fork, and delete are dispatched to a healthy connected executor rather than applied in Conductor's database, so they fail outright when nothing is running — resume and restart additionally need an executor on the latest version.
- **The autoscaling operations require a DBOS Teams plan** and return 403 otherwise.
- **Getting the spec from the Conductor image needs `--entrypoint`.** The image's entrypoint waits for Postgres and runs migrations without forwarding arguments, so the obvious `docker run <image> openapi` hangs waiting for a database the subcommand does not need.

## Two scoping calls worth a look

**The API is not "everything the console does."** An earlier draft said so; it is wrong, because the console also drives DBOS Cloud — deploying applications, provisioning databases, billing — none of which this API touches. Both pages now scope it to the Conductor control plane, which is what the console shows for a Conductor-connected application whether it runs on your own infrastructure or on DBOS Cloud, and point the cloud-specific operations at the pages that own them.

**Nothing is framed as versioned.** There is one public API; the predecessor is internal. `v2` appears only in URLs and paths, where you actually type it, and never in prose as though it were a choice.

## Verified against the shipped API

The pages track conductor `main` as of the `applicationName` revert (dbos-inc/dbos-conductor#187) and the CLI as of dbos-inc/dbos-ctl#8 and #9:

- the `listWorkflows` query parameters and the full workflow-search body field list match the spec exactly;
- all 64 operations appear in the endpoint tables, and all 16 OAuth-gated operations are named by operation ID;
- `applicationName` is documented as a **response** field only — the request filters were removed by #187;
- install instructions follow the `dbos-cli` → `dbos-ctl` repository rename and lead with the new install script.

The site builds. The only broken-anchor warnings are pre-existing ones on the Go reference pages, untouched by this branch.

## One sequencing note

The `dbosctl` page links to the CLI's releases page and its `raw.githubusercontent.com` install script, both of which are only reachable once that repository is public and a release is tagged. The page carries an early-access admonition, so it is not misleading in the meantime, but this probably wants to land alongside that flip.

🤖 Generated with [Claude Code](https://claude.com/claude-code)

### DIFF
--- a/docs/production/conductor-api.md
+++ b/docs/production/conductor-api.md
@@ -1,0 +1,321 @@
+---
+sidebar_position: 33
+title: Conductor API
+---
+
+Everything you can do from the [DBOS console](https://console.dbos.dev) you can also do over Conductor's HTTP API: list and manage applications, search workflows, cancel or fork them, inspect queues and schedules, drive schedules, read metrics and audit logs, and manage members, roles, and API keys.
+
+The API is versioned. This page documents **v2**, which is described by an OpenAPI 3.1 specification generated directly from the running server, so the spec is never out of date with the deployment serving it. The console and the [`dbosctl` CLI](./dbosctl.md) are both built on this API and on clients generated from that spec.
+
+## Base URL
+
+| Deployment | Base URL |
+| --- | --- |
+| DBOS-managed Conductor | `https://cloud.dbos.dev/conductor` |
+| [Self-hosted Conductor](./hosting-conductor.md) | `http://<your-conductor-host>:8090` (port `8090` by default) |
+
+Every v2 path is relative to that base, so the full URL of an operation is, for example:
+
+```
+https://cloud.dbos.dev/conductor/v2/orgs/my_org/apps/my-app/workflows
+```
+
+The paths themselves are identical in both deployments — only the base differs. This is deliberate: a single generated client works against DBOS-managed and self-hosted Conductor with nothing but the base URL changed.
+
+## The OpenAPI Specification
+
+There are three ways to obtain the spec.
+
+**From DBOS-managed Conductor.** The spec is served publicly (no authentication required) and reflects the currently deployed version:
+
+```shell
+curl -O https://cloud.dbos.dev/conductor/v2/openapi.json
+```
+
+If your toolchain does not yet support OpenAPI 3.1, request the 3.0 downgrade instead:
+
+```shell
+curl -O https://cloud.dbos.dev/conductor/v2/openapi-3.0.json
+```
+
+The spec served here is Conductor's own, with only its `servers` entry repointed at `/conductor` so that generated clients resolve paths correctly through DBOS Cloud. Nothing else — no path, schema, or security scheme — is rewritten or filtered.
+
+**From a self-hosted Conductor.** The server mounts the spec and an interactive browser at its root, all unauthenticated:
+
+| Path | Serves |
+| --- | --- |
+| `/openapi.json` | OpenAPI 3.1 specification (JSON) |
+| `/openapi.yaml` | The same specification in YAML |
+| `/openapi-3.0.json` | OpenAPI 3.0 downgrade |
+| `/docs` | Interactive API browser |
+| `/schemas/*` | The JSON Schema documents referenced by the spec |
+
+For example, with the Docker Compose setup from [Self-Hosting Conductor](./hosting-conductor.md), open `http://localhost:8090/docs` to explore the API in your browser.
+
+**From the Conductor binary.** The `openapi` subcommand prints the spec to stdout without connecting to a database or requiring any runtime configuration, which is convenient in CI and code generation pipelines:
+
+```shell
+conductor openapi > openapi.json
+conductor openapi -spec-version 3.0 > openapi-3.0.json
+```
+
+:::info
+The binary emits the *complete* route surface, including operations that a no-auth deployment does not register. See [Self-hosted differences](#self-hosted-differences) below.
+:::
+
+## Authentication
+
+All authenticated requests carry a bearer token:
+
+```
+Authorization: Bearer <token>
+```
+
+Conductor accepts two kinds of token, distinguished by their prefix:
+
+| Credential | Description |
+| --- | --- |
+| **API key** | A key beginning with `dbos_`, created with `POST /v2/orgs/{orgName}/tokens/{tokenName}` (or from the console, or with [`dbosctl api-key create`](./dbosctl.md#dbosctl-api-key-create)). Keys authenticate machine-to-machine callers and can be scoped to specific applications and permissions. A key carries no user identity, so it cannot call `GET /v2/users/me`. |
+| **User JWT** | An OIDC-issued JSON Web Token identifying a human user. This is what the console and `dbosctl login` use. |
+
+Both are sent the same way; Conductor tells them apart by the `dbos_` prefix.
+
+Authorization is enforced per operation using the permission model described in [Permissions and API Keys](./permissions.md) — a caller needs `application.read` to list workflows, `application.write` to cancel one, `organization.write` to manage members, and so on. An unauthenticated request returns `401`; an authenticated request lacking the required permission returns `403`.
+
+## Resource Naming
+
+Almost every operation is scoped to an organization, and most are additionally scoped to an application:
+
+```
+/v2/orgs/{orgName}/apps/{appName}/workflows/{workflowId}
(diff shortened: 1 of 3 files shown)


docs/production/dbosctl.md
```diff
@@ -0,0 +1,559 @@
+---
+sidebar_position: 34
+title: dbosctl CLI Reference
+---
+
+`dbosctl` is a command-line client for the [Conductor API](./conductor-api.md). It manages workflows, queues, schedules, applications, and API keys against DBOS-managed Conductor or a [self-hosted Conductor](./hosting-conductor.md), with the target selected by a named **profile**.
+
+:::info
+The binary is named `dbosctl`, not `dbos`. The DBOS language SDKs ship their own `dbos` entrypoints (the Python SDK, for example, installs a `dbos` console script), so the `ctl` suffix keeps this CLI unambiguous alongside any of them. `dbosctl` is also distinct from [`dbos-cloud`](./dbos-cloud/cloud-cli.md), which manages applications and databases hosted on DBOS Cloud.
```
kraftp: Not really needed, clear enough from the name of the page

docs/production/dbosctl.md
```diff
@@ -0,0 +1,559 @@
...
+dbosctl config set managed --managed   # create a profile pointing at cloud.dbos.dev
+dbosctl login                          # log in through the device-authorization flow
+dbosctl whoami                         # confirm who you are logged in as
+dbosctl app list
+```
+
+Against a self-hosted Conductor running without authentication:
+
+```shell
+dbosctl config set local --url http://localhost:8090
```
kraftp, line 60: What about self-hosted with auth?

docs/production/conductor-api.md
```diff
@@ -0,0 +1,334 @@
...
+| Bulk delete | `POST .../workflows/bulk-delete` |
+| Bulk fork from failure | `POST .../workflows/bulk-fork-from-failure` |
+
+The semantics of cancelling, resuming, and forking are described in [Workflow Management](./workflow-management.md). **Restart** starts a new execution of a workflow, with a new ID and the same inputs, from its first step; the original workflow is left untouched. The bulk variants take an array of workflow IDs and apply the same operation to each, which is far cheaper than issuing the calls one at a time. **Export** and **import** move a workflow and its steps between deployments as a JSON document — useful for reproducing a production failure in a development environment.
+
+:::info
+The mutating workflow operations — cancel, resume, restart, fork, and delete — are carried out by your application, not by Conductor's database. Conductor dispatches them over the websocket to a healthy connected executor. If an application has no healthy executor connected (resume and restart additionally require one running the application's latest version), the call fails rather than being queued for later.
+:::
+
+### Queues and autoscaling
```
kraftp: Let's not document this until it's done (I know it's in the OpenAPI spec, but we can keep it out of our public docs)

docs/production/conductor-api.md
```diff
@@ -0,0 +1,334 @@
...
+
+Workflow listing comes in two flavors:
+
+- **`GET .../workflows`** takes a few common filters as query parameters — `status`, `workflowName`, `limit`, `offset`, `sortDesc`, `loadInput`, `loadOutput` — and is convenient for quick queries.
+- **`POST .../workflows/search`** takes a JSON body and supports the full filter set the console uses: arrays of `status`, `workflowName`, `workflowIds`, `workflowIdPrefix`, `queueName`, `scheduleName`, `user`, `executorId`, `appVersion`, `parentWorkflowId`, and `forkedFrom`, plus `startTime`/`endTime`, `completedAfter`/`completedBefore`, `dequeuedAfter`/`dequeuedBefore`, `hasParent`, `wasForkedFrom`, `queuesOnly`, `attributes`, and the same paging and sorting fields.
+
+Workflow inputs and outputs can be large, so they are omitted unless you ask for them with `loadInput` and `loadOutput`.
+
+:::tip
+Several applications can share one system database. When they do, each workflow, queue, and schedule reports the application that owns it in an `applicationName` field, so you can tell whose objects you are looking at. The field is null for objects recorded before DBOS Transact tracked application names, and for in-memory queues.
```
kraftp: Probably remove this, API detail that doesn't deserve so much prominence (and misleading, `applicationName` is reported whether the system database is shared or not).

kraftp (approved):

docs/production/conductor-api.md
```diff
@@ -0,0 +1,334 @@
...
+curl -O https://cloud.dbos.dev/conductor/v2/openapi.json
+```
+
+If your toolchain does not yet support OpenAPI 3.1, request the 3.0 downgrade instead:
+
+```shell
+curl -O https://cloud.dbos.dev/conductor/v2/openapi-3.0.json
+```
+
+The spec served here is Conductor's own, with only its `servers` entry repointed at `/conductor` so that generated clients resolve paths correctly through DBOS Cloud. Nothing else — no path, schema, or security scheme — is rewritten or filtered.
```
kraftp:
```suggestion
The spec served here is Conductor's own, with only its `servers` entry repointed at `/conductor` so that generated clients resolve paths correctly through DBOS Cloud.
```

docs/production/conductor-api.md
```diff
@@ -0,0 +1,334 @@
...
+| Bulk resume | `POST .../workflows/bulk-resume` |
+| Bulk delete | `POST .../workflows/bulk-delete` |
+| Bulk fork from failure | `POST .../workflows/bulk-fork-from-failure` |
+
+The semantics of cancelling, resuming, and forking are described in [Workflow Management](./workflow-management.md). **Restart** starts a new execution of a workflow, with a new ID and the same inputs, from its first step; the original workflow is left untouched. The bulk variants take an array of workflow IDs and apply the same operation to each, which is far cheaper than issuing the calls one at a time. **Export** and **import** move a workflow and its steps between deployments as a JSON document — useful for reproducing a production failure in a development environment.
+
+:::info
+The mutating workflow operations — cancel, resume, restart, fork, and delete — are carried out by your application, not by Conductor's database. Conductor dispatches them over the websocket to a healthy connected executor. If an application has no healthy executor connected (resume and restart additionally require one running the application's latest version), the call fails rather than being queued for later.
```
kraftp:
```suggestion
Workflow operations are carried out by your application, not by Conductor's database. Conductor dispatches them over the websocket to a healthy connected executor. If an application has no healthy executor connected, the call fails.
```

docs/production/conductor-api.md
```diff
@@ -0,0 +1,334 @@
...
+| DBOS-managed Conductor | `https://cloud.dbos.dev/conductor` |
+| [Self-hosted Conductor](./hosting-conductor.md) | `http://<your-conductor-host>:8090` (port `8090` by default) |
+
+Every path is relative to that base, so the full URL of an operation is, for example:
+
+```
+https://cloud.dbos.dev/conductor/v2/orgs/my_org/apps/my-app/workflows
+```
+
+The paths themselves are identical in both deployments — only the base differs. This is deliberate: a single generated client works against DBOS-managed and self-hosted Conductor with nothing but the base URL changed.
```
kraftp:
```suggestion
The paths themselves are identical in both deployments; only the base differs.
```

PR #612, opened 2026-08-21
Quickstart Overhaul

kraftp:


docs/quickstart.md
```diff
@@ -120,8 +120,50 @@ Congratulations, you've run your first durable workflow with DBOS!
...
+
+<article className="col col--6">
+```bash
+dbosctl config set dbos --managed
+dbosctl login
+```
+</article>
+
+<article className="col col--6">
+Then, register your application with Conductor and create an API key.
```
qianl15, line 153: Do we want to clarify that the API key can be reused for multiple apps?

qianl15 (approved):

docs/quickstart.md
kraftp: Not needed in the quickstart, I think (already clarified in docs)

